PR #4, opened 2015-10-20
Move react dependency to peerDependencies

damacisaac:
In order to maintain compatibility with projects using react v0.14.0, we want to use the react version installed on the owner's project. This also prevents a separate version of react being packaged with react-markdown.


natew:
:+1:

rexxars:
Thanks!

PR #15, opened 2016-02-14
Allow walker callback

firsara:
allows manipulating child nodes before rendering.
Useful if nodes need to be processed in specific scenarios.


rexxars:
Thanks for contributing!

I don't think the solution regarding the "double walking" is the right choice. The event has both an  `entering` property and an `isContainer` property. With these two, users should be able to figure out for themselves if they want want to process the node on entering, leaving, or both.

So, if you could pass the event instead of just the node and remove the `didWalk`-logic, I'll merge this :+1:

firsara:
Oh didn't know about those properties, thanks for mentioning. I was just wondering about the double calls when I was implementing it.

Changed it, thank you (y)

rexxars:
Thanks again! I have a few more changes I want to get into master, then I'll push out a new release

firsara:
Nice. Great module btw :)

PR #24, opened 2016-04-19
Add `childBefore` and `childAfter` props

thomaslindstrom:
In some cases, I'd find it useful to render children *before* and/or *after* the rendered markdown, yet *within the same container element*.

**Simple pseudo example:**
```
var childBefore = (<img src="avatar.png" alt="User avatar" />);
<ReactMarkdown source="**Hello**" containerTagName="div" childBefore={childBefore} />
```
```
<div>
   <img src="avatar.png" alt="User avatar" />
   <strong>Hello</strong>
</div>
```


rexxars:
Thanks!

PR #28, opened 2016-06-06
Pass the walker to the walker callback function

rikukissa:
This comes in handy when you need to do destructive updates for the AST and have to call `walker.resumeAt` afterwards.


rexxars:
Thanks!

PR #41, opened 2016-11-01
Add TypeScript definitions.

IRus:
Hi, this is definitions for TypeScript.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,13 +54,13 @@
     "staticpath": "demo/dist"
   },
   "dependencies": {
+    "@types/react": "*",
     "commonmark": "^0.24.0",
     "commonmark-react-renderer": "^4.2.4",
     "in-publish": "^2.0.0"
   },
   "types": "./react-markdown.d.ts",
   "peerDependencies": {
-    "@types/react": "*",
     "react": ">=0.13.3"
   }
 }
--- a/package.json
+++ b/package.json
@@ -58,7 +58,9 @@
     "commonmark-react-renderer": "^4.2.4",
     "in-publish": "^2.0.0"
   },
+  "types": "./react-markdown.d.ts",
   "peerDependencies": {
+    "@types/react": "*",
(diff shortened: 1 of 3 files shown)


IRus:
@rexxars hi! any concerns about this PR?

rexxars:
Hi, sorry for taking time to look at this. Is it common practice to have a peer dependency on the react typedep? It seems like this will lead to a lot of warnings during installation for anyone not using typescript?

IRus:
Hi! I'm really not sure. I found two common ways to deal with `types`:
1. Define as `dependency` with version `*`.
2. Define as `peerDependency` with version `*`.

Types should be "singleton" in node_modules, because two versions of types lead to error (typescript told about duplicate definitions)

Also i looked at usages of `@types/react`: https://www.npmjs.com/browse/depended/@types/react
And there are a lot of such usages: https://github.com/danielearwicker/immuto-react/blob/master/package.json#L28

And this is wrong for sure.

IRus:
@rexxars I moved `@types/react` to deps.

rexxars:
Sorry for not getting back to you sooner, but I'm not sure if this makes it any better...
Axios has a [typings.json](https://github.com/mzabriskie/axios/blob/master/typings.json) file which seems to declare something along the same lines - is this the way to do this?

Sorry for being pedantic here, I just don't want:
* Unnecessary dependencies for people not using typescript
* Warnings on installation for people not using typescript

IRus:
Hi again! Typings and tsd are deprecated after typescript 2.0 release. 

Okay, I got your point. I think most people that uses typescript already installed `@types/react` before installing your package there is no problem to delete dependency to `@types/react`.

If you like, i can move it to devDeps, so then example  in typescript will work properly in development.

rexxars:
Sure, devDeps works for me!

IRus:
Done!

rexxars:
Thanks!

PR #60, opened 2017-03-02
fix: Too strict TS type

### LINKED ISSUE #59: Too strict TS typings
The renderers prop is typed as `readonly renderers?: {[nodeType: string]: Component<any, any>};`
however `Component` only allows class component, not functional components. 

The react typings does instead export a type named `ReactType` that can be used for all components. So the renderers prop should be typed as  `readonly renderers?: {[nodeType: string]: ReactType};`

Pajn:
fixes #59


rexxars:
Thanks!

PR #63, opened 2017-03-14
Update Installation Instructions

rwieruch:
- fixing "It's no longer allowed to omit the '-loader' suffix when using loaders"


rexxars:
Thanks!

PR #66, opened 2017-03-22
Remove `React.createClass`

goto-bus-stop:
Hi! :wave: Thanks for creating this component!

The `React.createClass` method is going to be deprecated soon and will
be removed in React v16.
https://github.com/facebook/react/pull/9232
    
This patch changes the component to a class extended from
`React.Component`. It does it manually by assigning to the prototype,
to avoid introducing a compile step.

e; Oops, goofed up the title there with `hub pull-request` :sweat_smile:


rexxars:
Thanks for the contribution! Appreciated.

PR #71, opened 2017-04-08
Fix some more deprecations for React v15.5

goto-bus-stop:
Hi again! :wave: This one comes several parts:

 - Use `prop-types` module instead of `React.PropTypes`, see https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.proptypes
 - Use `react-dom/test-utils` instead of `react-addons-test-utils`, see https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#react-test-utils
 - Also update the demo so it doesn't use deprecated stuff anymore:

   Adds `buble-loader` to compile ES6 classes that extend from
   `React.Component` to ES5, in the demo only. I figured since there
   was already a Webpack build for the demo, adding a small extra
   step wouldn't hurt :)

   Also removes the `pure-render-mixin` dependency, because that's
   only useful with `createClass`.


rexxars:
Great work! Thanks!

rexxars:
Released as 2.5.0

goto-bus-stop:
Sweet! Thank you! :tada:

PR #106, opened 2017-11-21
transformLinkUri - provide all available arguments

mudrz:
Currently `transformLinkUri` can be provided for custom URL transformations, but not all available arguments are passed.
This makes it impossible to provide custom URL transformation logic, based on the URL title and content.

With this change both children and title are provided to the `transformLinkUri` (without changing the default `uriTransformer` implementation


rexxars:
Thanks!

rexxars:
Released as 3.0.1.

mudrz:
nice, you are fast 👍

PR #121, opened 2017-12-08
Adding readonly to checkboxes. Fixes #120.

### LINKED ISSUE #120: Clicking checkboxes
Hi,

If I use source like `- [ ] checkbox` and click on the checkbox, this warning is displayed:
```
Warning: Failed prop type: You provided a `checked` prop to a form field without an
`onChange` handler. This will render a read-only field. If the field should be mutable
use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`.
    in input (created by ListItem)
```
It would be interesting to provide some way to actually handle the state changes other than changing the markdown source itself, but it sounds like a lot of work to handle the fact that the structure of the checkbox inputs is completely dynamic depending on the source. So it might be a good idea to make checkbox inputs `readOnly` by default to at least suppress the warning and make the behavior clear.

philraj:


rexxars:
👍

rexxars:
Released as 3.1.1

PR #130, opened 2017-12-27
enable transformImageUri also for image references, not only direct images

evoye:
currently, the transform function is not yet applied for referenced images


rexxars:
Thanks!

PR #144, opened 2018-02-20
Pass the React Key into an overridden text renderer and update the README

vanchagreen:
## Commit 1
If you override the text renderer with a function that returns a React Element, you'll get a "unique key error" warning if you don't pass in a key.

This commit passes the key that `astToReact` gives to every other node to the text renderer function. If you are simply returning a string from your renderer function, you can just ignore the extra parameter.

Note: another way to fix this would be to have the text renderer accept a React Component like the rest of the renderers. That would be a breaking change at this point though.

## Commit 2
The current README docs are slightly wrong.

When you override the `text` renderer it expects a function, but the docs imply that it always takes a React Component. If you pass it a React Component you will get an error.

This PR lets users know that they should pass in a function instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ this has been planned, but if you're feeling up to the task, create an issue and
 * `renderers` - _object_ An object where the keys represent the node type and the value is a React
   component. The object is merged with the default renderers. The props passed to the component
   varies based on the type of node.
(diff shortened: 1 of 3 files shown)


rexxars:
Thanks a bunch. Released as v3.2.1!

PR #150, opened 2018-03-05
Add support for fragment renderers

bsonntag:
This PR updatest `astToReact` so that it supports passing `React.Fragment` as a renderer without wrapping it in a function.


rexxars:
Thanks a bunch! Released as 3.3.0 :)

PR #161, opened 2018-03-31
Update TypeScript definition and prop types

ClassicDarkChocolate:
1. Match the unified API use `unified().use(params1).use(params2)...` with `plugins: [ [ params1 ], [ params2 ], ... ]`
2. Support for `remark-iframes` options. Example usage: `plugins: [ remarkIframes, { 'www.youtube.com': {...} } ]`
3. Update TypeScript definition.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for react-markdown v3.0.0-rc1
+// Type definitions for react-markdown > v3.3.0
 // Project: https://github.com/rexxars/react-markdown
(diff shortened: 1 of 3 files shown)


rexxars:
1. Already implemented, there was no changes in this PR related to this apart from using spread instead of indexed lookups, but accepting those changes as it reads slightly better
2. #170 solved this in a more generic way, passing on node properties for unknown nodes directly to renderers.
3. Thanks a bunch! Accepted these changes and reworded the PR as it reads better in the commit history.

PR #152, opened 2018-03-09
Pass node properties to renderer

juangl:
This allow us to have an abstract representation of custom nodes and passing props to the render function such as:

```javascript
{
  'type': 'linkbar',
  'description': 'some text',
  'image': 'image.jpg',
}
```

and render them as:

```javascript
renderLinbar = ({description, image}) => 
  <Linkbar description={description} image={image} />
```

PR #170, opened 2018-05-07
Pass node properties to renderer

jessepinho:
This is essentially a duplicate of #152, but with a test added.

@rexxars, what do you think of the way I did the test? I included `remark-shortcodes` as a dev dependency, for the purpose of verifying that its attributes get passed along.

The problem is, when I run the tests, it says no assertion calls are being made. Furthermore, calling `console.log(renderHTML(component))` at the end of my test shows that it's being treated as a link reference—as though the passed-in `shortcodes` plugin is having no effect. Any chance you could help with this? Not sure where I'm going wrong.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "react-dom": "^16.2.0",
     "react-test-renderer": "^16.2.0",
     "remark-breaks": "^1.0.0",
+    "remark-shortcodes": "^0.1.5",
     "rimraf": "^2.6.2",
     "webpack": "^4.1.0",
     "webpack-cli": "^2.0.10"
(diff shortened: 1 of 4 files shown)


rexxars:
The shortcodes syntax clashes with the linkDefinition syntax, or it can't be used inline. I'm not sure -  not familiar with shortcodes. I replaced the shortcode example used in the tests with something on separate lines, which was correctly parsed and treated as shortcode.

Tests are passing now. Thanks for the effort!

jessepinho:
Great! Thanks so much @rexxars  :) It works!

PR #177, opened 2018-05-28
Remove todo from HTML renderer

lucasecdb:
Fiber landed, so there's no need for the call to `createElement`.


rexxars:
Thanks, but I'm not ready to drop support for React 15 yet.

PR #182, opened 2018-06-14
Fix bug with some disallowed nodes

petrgazarov:
This fixes a scenario when a disallowed node does not have children. Current code will evaluate `node.children` and add undefined to  `parent.children`'s array. This should not happen because the consumer code relies on a child not being `undefined`.

This problem manifests only when mode is `unwrap`.

An example of child with this problem is `thematicBreak` (`source: "***"`)


rexxars:
Thanks a bunch! Released as v3.3.3

PR #185, opened 2018-06-19
FIX: parsing empty link references

denis-s-svsg:
Fix bug when in markdown template presents empty link reference `[][]`

For such values at time of mapping syntax tree to react components in this line
https://github.com/rexxars/react-markdown/blob/master/src/ast-to-react.js#L57
`node. identifier` is empty string and as result `ref` will be `null`.  As result later in line 
https://github.com/rexxars/react-markdown/blob/master/src/ast-to-react.js#L105
at time of getting `ref.href` application will throw exception.

Fixed it. Test case added.

### DIFF
--- a/src/ast-to-react.js
+++ b/src/ast-to-react.js
@@ -54,7 +54,7 @@ function getNodeProps(node, key, opts, renderer, parent, index) {
     props.sourcePosition = node.position
   }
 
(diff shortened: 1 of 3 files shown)


rexxars:
Fantastic, thanks! Released as v3.3.4

PR #200, opened 2018-07-23
Updated ReactMarkdownProps to prevent forcing of source attribute

gRoberts84:


rexxars:
Thanks! Published as 3.4.0

Issue #146, opened 2018-02-22
Feature request: Set default target for links


GoodGuyMarco:
Is it possible to render a link with target="_blank" attribute without using a custom renderer?

rexxars:
Not at the moment. Based on the number of requests for this I'm thinking it would probably be wise to just add an option for it, though.

GoodGuyMarco:
@rexxars I would really appreciate that! Could you provide an example of a custom renderer which is capable of that until now?

rexxars:
```js
<Markdown
  source={yourMarkdownSource}
  renderers={{link : props => <a href={props.href} target="_blank">{props.children}</a>}}
/>
```

joshkadis:
You could also make the `target` attribute optional by appending a string to the `href`, like:
```
// Try something like https://example.com{_blank}
export default ({ href, children }) => {
  if (/{_blank}$/.test(href)) {
    return <a
      href={href.replace('{_blank}', '')}
      target="_blank"
    >
      {children}
    </a>;
  }
  return <a href={href}>{children}</a>;
};
```

rexxars:
Marshall Smith (@marshallds) added a `linkTarget` prop that now lets you do this without a custom renderer. This was released in 3.5.0, see the README for details.

PR #210, opened 2018-08-21
Add target attribute to links [#146]

marshallds:
Addresses issue #146 by adding a `linkTarget` attribute to default the target attribute for all links.

This can also be set to a function to support the case where links to specific domains should open in new tabs but not other links.

I considered the url parsing strategy suggested in #146 but did not implement it as it's non-standard markdown. I believe a better approach would be to use a mdast plugin.


rexxars:
Released in 3.5.0

PR #211, opened 2018-08-21
Add support for passing options to remark-parse plugin

RouganStriker:
Will also address https://github.com/rexxars/react-markdown/issues/129

Not sure if the test is good since it is dependent on a specific behaviour in `remark-parse`.


rexxars:
Real sorry about not following this up - I've manually merged this and changed the property name to `parserOptions` - released in v4.1.0 - thanks!

PR #214, opened 2018-09-04
Add support for passing index info to renderers

### LINKED ISSUE #213: Passing index information to renderers
Hello! Thank you *so much* for creating and maintaining such an awesome library! 👏 

I'd like to be able to conditionally apply some styles based on a node's index under its parent, e.g. to remove some top margin from the first element and the bottom margin from the last element. Basically enabling some of the stuff that in CSS-land we would have done with `:first-child`, `:last-child`, `:nth-child`, etc.

If you're open to it, I have a PR that mirrors what you did with `rawSourcePos`, passing index information to non-tag renderers if a boolean prop on the `Markdown` component is set to `true`. Would love your feedback! And I'm particularly terrible at naming things so please feel free to suggest better alternatives!

Will open that PR momentarily.

beauroberts:
When `indexUnderParent` prop is `true`, pass `indexUnderParent` and `parentChildCount` props to non-tag renderers.

Resolves #213

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ this has been planned, but if you're feeling up to the task, create an issue and
 * `sourcePos` - _boolean_ Setting to `true` will add `data-sourcepos` attributes to all elements,
   indicating where in the markdown source they were rendered from (default: `false`).
 * `rawSourcePos` - _boolean_ Setting to `true` will pass a `sourcePosition` property to all renderers with structured source position information (default: `false`).
+* `indexUnderParent` - _boolean_ Setting to `true` will pass `indexUnderParent` and `parentChildCount` props to all renderers (default: `false`).
 * `allowedTypes` - _array_ Defines which types of nodes should be allowed (rendered). (default: all
   types).
 * `disallowedTypes` - _array_ Defines which types of nodes should be disallowed (not rendered).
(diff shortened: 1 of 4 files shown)


rexxars:
I like the idea, but not crazy about the `indexUnderParent` name, perhaps we could just name it `index`? It's already being passed down for list items - I suspect we could pass it for all node types.

rexxars:
For a list that starts at say, 2, the `index` is the actual index (from zero), while the parent `list` node will receive a `start` property of `2`. So I think it should be safe. I would generally just drop the prop exposed on the markdown component and always pass those props to the child renderers.

As for the `parentChildCount`, I can't come up with anything more descriptive there. Open to suggestions, but I don't have a big issue with keeping it as-is.

rexxars:
Fair point. I think for a 4.0 I would want to change this a bit - not being able to pass down new properties without breaking peoples code seems very limiting. Let's put them behind a bool prop for now and I can make it the default behavior in 4.0.

How about `includeNodeIndex` as the prop name?

beauroberts:
Ready for another review @rexxars!

rexxars:
Thanks again, merged and released in 3.6.0

PR #219, opened 2018-09-25
Strongly type the keys of `renderers`

LinusU:
Just had a problem where I passed in `{ Link: ... }` instead of `{ link: ... }`, having this change would have caught it!


rexxars:
@LinusU Are you sure this change is correct? I am getting errors from VS Code when looking at the typescript definition: "An index signature parameter type cannot be a union type"

Should this be `readonly renderers?: {[nodeType in NodeType]: ReactType}`?

Issue #221, opened 2018-09-26
Links don't work if any characters in 'https' are uppercase


liamsain:
Hi. 
When trying to render this: `[link](Https://www.google.com)`, the href value comes through as 'javascript:void(0)'. same case for 'hTtps', 'htTps' etc, as well as all caps.

Any ideas for fixes /workarounds?

Thanks

liamsain:
pr: https://github.com/rexxars/react-markdown/pull/222

rexxars:
Thanks! Scheduled for 4.0

PR #222, opened 2018-09-26
converting urls to lower case before checking whether they contain a …

liamsain:
…protocol

fix for issue 'Links don't work if any characters in 'https' are uppercase #221'


rexxars:
Thanks! Scheduled for 4.0

PR #232, opened 2018-10-11
Add rendering of blockquote so that line breaks are properly rendered

### LINKED ISSUE #230: Render newlines as <br> in blockquote?
Currently it just prints a newline character, so you would have to use hard break or CSS to actually wrap where expected.

tias79:
Fixes #230


wooorm:
This breaks users who expect markdown to work as markdown

Issue #247, opened 2018-11-23
More documentation related to renderers


annjawn:
Hi, great library. I am having a hard time understanding the usage of renderers. Is there a way to add some more documentation with small examples on how to use it? Basically my use case is pretty simple, i have have some markdown text in a `.md` file and I use Semantic UI React for UI components and I want to be able to use some of the Semantic UI components as renderers (like buttons etc.). Is that even possible?

annjawn:
Nevermind.... got it working.. Thanks.

el1f:
I think this is still an on point observation though. Some renderers provide extra props that aren't documented and that needs us to go look at the vanilla renderers implementation to see which props are needed by which renderer... a table in the readme.md would do a great job for that

annjawn:
The way I did it was by `console.log` the props and looking at source code. Not ideal, but yes some more documentation on how to actually use the renderers would be very helpful specially since different renderers have different props that can be accessed. I also think there can be a few more default node types that can be supported like `<details>`.

PR #251, opened 2018-11-28
Added better documentation for renderer props

el1f:
Do you think there's a better way to display this kind of information?
References #247

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Should be straightforward. You might need to alter you code slightly if you:
 - Rely on there being a container `<div>` without a class name around your rendered markdown
 - Have implemented a custom `text` renderer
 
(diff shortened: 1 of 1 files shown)


rexxars:
Adding props is a nice touch, but I don't really like how this looks in a table. The description takes up most of the width and the props are hard to read jammed at the end. It also doesn't scale very well on smaller (mobile) devices.

Don't have a clear picture of what a better solution would be right now. Will give it some thought.

el1f:
You have a point. Gonna try with another configuration. Maybe having a list with `code` props on a new line could help

el1f:
Tried doing it with a separate renderer props section, with links this could prove as a less hard to approach implementation

SalahAdDin:
Good!

PR #254, opened 2018-12-05
Allow arbitrary renderer types

jessepinho:
Since [renderers for arbitrary node types are allowed](https://github.com/rexxars/react-markdown/issues/162), the typings should reflect this.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -79,7 +79,7 @@ declare namespace ReactMarkdown {
     readonly transformLinkUri?: (uri: string, children?: ReactNode, title?: string) => string
     readonly transformImageUri?: (uri: string, children?: ReactNode, title?: string, alt?: string) => string
     readonly unwrapDisallowed?: boolean
-    readonly renderers?: {[nodeType in NodeType]?: ReactType}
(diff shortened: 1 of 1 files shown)


ihsw:
Can this be merged? It would be wonderful.

jessepinho:
Will fix ASAP!

jessepinho:
@ihsw I'm not seeing any build errors locally. Note that there was already a `?` there in the original code — I only change the type of the `nodeType` key, not whether it was optional. 

Can you please confirm whether this is still an issue, or if it can just be merged? Thanks!

rexxars:
The `?` was there because the old version specified that the key had to be one of the `NodeType` values, but that you _did not have to specify all of them_. With this change, the question mark is (as far as I know) unnecessary because any string is now a valid key.

ihsw:
@jessepinho attached is a screenshot.

2019-01-03-085355_1484x755_scrot

TypeScript 3.2.2

jessepinho:
Ah, I understand! Just pushed up the requested change — should work now. Thanks!

PR #260, opened 2019-01-03
Upgrade remark-parse

rgbkrk:
Unsure if this will break things on the `react-markdown` side. I wanted to see about upgrading remark (xref: https://github.com/nteract/nteract/pull/3483).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8049,9 +8049,9 @@
       "dev": true
     },
     "remark-parse": {
-      "version": "5.0.0",
-      "resolved": "https://registry.npmjs.org/remark-parse/-/remark-parse-5.0.0.tgz",
-      "integrity": "sha512-b3iXszZLH1TLoyUzrATcTQUZrwNl1rE70rVdSruJFlDaJ9z5aMkhrG43Pp68OgfHndL/ADz6V69Zow8cTQu+JA==",
+      "version": "6.0.3",
+      "resolved": "https://registry.npmjs.org/remark-parse/-/remark-parse-6.0.3.tgz",
+      "integrity": "sha512-QbDXWN4HfKTUC0hHa4teU463KclLAnwpn/FBn87j9cKYJWWawbiLgMfP2Q4XwhxxuuuOxHlw+pSN0OKuJwyVvg==",
       "requires": {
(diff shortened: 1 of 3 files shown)


rgbkrk:
Wow, not a fan of the markup that `remark-parse` is now generating for that unordered list:

```
    Received value does not match stored snapshot "should handle loose, unordered lists 1".
    - Snapshot
    + Received
      <ul>
        <li>
    -     <p>
    -       foo
    -     </p>
    +     foo
        </li>
        <li>
    -     <p>
    -       bar
    -     </p>
    +     bar
        </li>
      </ul>
```

rgbkrk:
@rexxars would you have a chance to check this out? I upgraded `remark-parse` and updated the snapshot ✨ .

rgbkrk:
Updating the `package-lock.json` now.

rgbkrk:
Hurrah! It passes!

rexxars:
This is a breaking change, since we are no longer getting a `loose` property from remark.

Remark has changed their [list handling](https://github.com/remarkjs/remark/pull/364) and thus we need to reflect this in our renderers by doing similar to [mdast-util-to-hast](https://github.com/syntax-tree/mdast-util-to-hast/commit/06361e223902d5557cbd56f38cbaad83a2650ccb)

This PR currently just handles all lists as if they were "tight", which isn't necessarily what people want.

rgbkrk:
Darn. Beyond the other changes necessary, does this mean you'd want to make a major version bump or forgo upgrading remark?

rexxars:
I don't really mind a breaking change if it makes sense, but as far as I can tell the functionality is still possible to reproduce with the latest remark, it just needs to be handled differently. I definitely don't want to remove a feature if it's supported by the underlying stack.

rgbkrk:
I'm not sure what the next steps I should take are, so pending further direction I think I'll have to leave this to other contributors.

rexxars:
Thanks for the PR, I'm scheduling the upgrade to remark-parse@6 for v5, which is in a separate branch right now: https://github.com/rexxars/react-markdown/tree/v5

rgbkrk:
Cool, looking forward to it.

PR #261, opened 2019-01-05
Typings: use Renderers for renderers of ReactMarkdownProps

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
I think `renderers` of `ReactMarkdownProps` should use interface `Renderers`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -79,7 +79,7 @@ declare namespace ReactMarkdown {
     readonly transformLinkUri?: (uri: string, children?: ReactNode, title?: string) => string
     readonly transformImageUri?: (uri: string, children?: ReactNode, title?: string, alt?: string) => string
     readonly unwrapDisallowed?: boolean
-    readonly renderers?: {[nodeType: string]: ReactType}
+    readonly renderers?: Renderers
     readonly astPlugins?: MdastPlugin[]
     readonly plugins?: any[] | (() => void)
   }


rexxars:
Are there any specific reason why that would be better? I am not super familiar with typescript yet.

There has been a lot of changes in this lately, maybe @jessepinho or @ihsw has any input?

jessepinho:
@b6pzeusbc54tvhw5jgpyw8pwz2x6gs Is [`Renderer<T>`](https://github.com/rexxars/react-markdown/blob/c25a51660a35f3a7b3d58e2a6b6eeffde91fe0a9/index.d.ts#L91) roughly the same as `ReactType`? If so, this should be fine. Also, have you tested your change?

ihsw:
@rexxars it could be better as it more clearly conveys that the return type can be a `string` or `Renderer`.

`Renderer<T>` is roughly the same as `ReactType<T>` though.

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@rexxars @jessepinho @ihsw 
Thank you for your discussion. I am also not familiar with typescript yet.

The [builtin default renderers.js][renderes] has **string** return type. I think the default renderers and user customized renderers both should use the same type.

[renderes]: https://github.com/rexxars/react-markdown/blob/7cddebfeb1891fdcceff3571f65dabe53241e146/src/renderers.js

jessepinho:
Again, have you tested your change on a custom renderer? (And you mentioned you're not familiar with TypeScript — are you actually using this change? If not, let's not change something that isn't broken. 😄 )

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@jessepinho Yes, I tested my change.

It is my custom renderers,
```typescript
const customRenderers: ReactMarkdown.Renderers = {
  linkReference: (props) => {
    const videoUrl = props.href
    const thumbnailUrl = props.children[0].props.src
    if (
      (lookupMime(videoUrl) || '').split('/')[0] === 'video'
      && (lookupMime(thumbnailUrl) || '').split('/')[0] === 'image'
    ) {
      // we use custom video player on this case
      return <span style={{ display: 'none' }}/>
    }
    return <a {...props}/>
  },
}
```

Before the change, I encountered the following error, but after the change, the error disappears.
screen shot 2019-02-06 at 1 15 13 am

Rocamonde:
Dear all, @rexxars 

What's going on with this? This still seems to be an issue today. I cannot seem to use this project with TypeScript. Is this fix planned for any upcoming release?

Best

wooorm:
Hi, thanks for your patience!

PR #292, opened 2019-04-08
Add support to web workers

fcostarodrigo:
It worked fine with a create-react-app test that I made, the parsing is done in the web worker in parallel with the animation, but there is a moment that the animation still freezes. And unfortunately my project is still on webpack 3 and the worker loader needs version 4. I will try something else here, but I will leave the pull request.

It appears it is not possible to pass the list of plugins to the web worker, since you can't pass functions to it. https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,24 @@ const parseHtml = htmlParser({
 />
 ```
 
+## Using web workers
+
+If you want to do the parsing in web workers, import `react-markdown/with-web-worker` instead of the default:
+
+```js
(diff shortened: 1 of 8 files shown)


rexxars:
Thanks for giving it a shot! I'm afraid I don't want to maintain this in core - it's just too big a departure right now. Obviously, feel free to maintain this as a fork if you want.

PR #296, opened 2019-04-14
Fix matching of replaced non-void elements in HTML parser plugin

nicolasv:
In `parseNode()` from `html-parser.js` we first match against `closingTagRe` to push onto `open` to indicate we're in a non-void element.

However, if on the matching opening tag you return an element — from `parser.parseWithInstructions()` — with a different tag as what was matched by `closingTagRe`, or a custom React component (where `type` is no longer a string), you can't close off the matched non-void element.


rexxars:
Good catch. Thanks!

rexxars:
Released as v4.0.7

PR #324, opened 2019-07-22
putting [] [] in the source breaks

Axosoft-Ashley:
Putting [] [] in the source breaks, because ref is null then we dont null check ref.href below.


rexxars:
Sorry, I'm not able to reproduce this... Do you have some sample markdown that will make it crash? I want to make sure we have a test case that covers this before I merge.

Axosoft-Ashley:
Hello! Thanks for the response! I found we were using an older version of markdown and updated it. Seems like this commit fixed it for me:

https://github.com/rexxars/react-markdown/pull/185

Thanks!

PR #335, opened 2019-09-02
Fix for remark-parse/types has no exported member RemarkParserOptions

hsz:
WIth https://github.com/rexxars/react-markdown/commit/fe0b57a8e8e7256446419e2d49064a78c6d19479 commit, there was introduced `parserOptions` property with `RemarkParserOptions` type.

There is typo in the interface name - should be: `RemarkParseOptions`


rexxars:
Thanks, will do a patch release.

hsz:
Thanks for the quick response and release!

PR #341, opened 2019-09-17
Better type definitions for `renderers` prop

danilofuchs:
Currently, `renderers` prop object keys are expanded to string, which can cause bugs if the user types the wrong NodeType as a key.
Using `Record<NodeType, ReactType>`, the object's keys must be valid `NodeType`. `Partial` is needed to allow for partial renderer override.


kiliancs:
See #417 and #417

wooorm:
Hi, thanks for your patience!

Issue #283, opened 2019-03-14
Add 'data' as valid URI protocol


fxh1357:
I would like to render anchor tags with [data-URIs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs). Right now links passed via source property will be sanitized by react-markdown/src/uri-transformer.js to match one of the predefined protocols 'http', 'https', 'mailto' or 'tel'. 

Adding 'data' to the collection would allow me to render asynchronously loaded files as links inside of localized markdown.

rexxars:
Sure - would happily accept a pull request for this.

juancarlosqr:
You got a PR for this!

MajorSaheb:
please add other uri too..
sms: 
skype:
whatsapp: 
market:
@rexxars kindly look into it.

c7hudson:
+1 for SMS capabilities.

juancarlosqr:
Hey @rexxars have you looked into this one? Please provide your opinion on adding more protocols as requested, thanks. I can gladly add them

ChristianMurphy:
this will be supported in #428

wooorm:
Not going to happen by default. It introduces serious security issues ([e.g., this](https://snyk.io/vuln/npm:markdown-it:20160912)). And you can already live dangerously with the `transformLinkUri` option.

PR #349, opened 2019-10-05
Add data to allowed uri protocols

juancarlosqr:
Solves issue #283 😉


c7hudson:
Bump - might be looking to open similar PR with added to protocols.

wooorm:
See https://github.com/remarkjs/react-markdown/issues/283#issuecomment-710027246

PR #355, opened 2019-10-20
WIP: add more strict typing for renderers

ivan-nikolaievskyi:
Using this library in my project, I found out that TS typing for renderers check almost nothing.
You can add the wrong key or don't get what the list of properties will get rendered until you debug it.
This problem makes writing renderers more complicated and buggy.
In this PR I tried to solve this problem.
This is not production-ready and finished implementation. 
It's just an example of how to add more strict typing for renderers.
Please check this one. I can finish it if needed or you can build your own definition based on that. 
Thanks.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -32,33 +32,7 @@ declare namespace ReactMarkdown {
     pedantic: boolean
   }
 
-  export type NodeType =
-    | 'root'
-    | 'text'
-    | 'break'
-    | 'paragraph'
-    | 'emphasis'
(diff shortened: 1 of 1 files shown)


split:
What kind of work this still needs? These types seem to be exactly what defining custom renderers in typescript seem to need. Currently renderer types are so broken you can't even pass default renderers as props to Markdown component.

Would be awesome to get this in 👍

wooorm:
Hi, thanks for your patience!

PR #356, opened 2019-10-25
Fix React warning for sourcePosition prop

bradbyte:
Hello,

Thanks so much for maintaining this awesome library! 

I started using the `rawSourcePos` prop and noticed a React warning about a DOM property being set for `sourcePosition`. I traced the error to the `root` renderer and found it wasn't using the `getCoreProps` cleaning method. I updated the renderer to use it and still allow for `className` to be set. Also added a test to ensure the root doesn't render `sourcePosition` props (the test failed before my fix was in place).

Here's the warning I was getting...
Screen Shot 2019-10-24 at 11 38 47 PM

### DIFF
--- a/src/renderers.js
+++ b/src/renderers.js
@@ -45,7 +45,11 @@ function TextRenderer(props) {
 function Root(props) {
   const useFragment = !props.className
   const root = useFragment ? React.Fragment || 'div' : 'div'
-  return createElement(root, useFragment ? null : props, props.children)
+  const rootProps = Object.assign(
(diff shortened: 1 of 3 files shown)


wooorm:
Hi there, thanks for your patience!

This looks good to me! Could you point this PR to the v5 branch? https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request and fix merge conflicts if there are any?

PR #365, opened 2019-11-13
Always wrap spans with text even in React 16 to deal with Google Translate

erulabs:
Thanks for React-Markdown! I've noticed users with Google Translate enabled run into a few issues adding/removing dom elements - related to this bug: https://github.com/facebook/react/issues/13278

The errors are like `Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.` and `Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.`

It appears to be related to the `TextRenderer` element simply returning string literals. While this does work fine in React 16 (seems to be a performance improvement attempt?), it breaks when users have Google Translate enabled. The TLDR is that the string literal is wrapped in a `font` element, silently on the DOM,  which causes React to lose its reference. The "fix" is to always wrap string literals in known-to-react dom elements - thus, always wrapping text with `spans` resolves the problem and fixes React-Markdown for users with Google Translate enabled.

Hopefully this helps! If there is some context around the `supportsStringRender` that I might be missing, please let me know and I can fix the PR!

Thanks!

### DIFF
--- a/src/renderers.js
+++ b/src/renderers.js
@@ -4,7 +4,6 @@
 const xtend = require('xtend')
 const React = require('react')
 
-const supportsStringRender = parseInt((React.version || '16').slice(0, 2), 10) >= 16
 const createElement = React.createElement
 
 module.exports = {
@@ -39,7 +38,7 @@ module.exports = {
 }
 
 function TextRenderer(props) {
-  return supportsStringRender ? props.children : createElement('span', null, props.children)
+  return createElement('span', null, props.children)
 }
 
 function Root(props) {


erulabs:
Oh boy this breaks a lot of tests. Let me know if this is a reasonable patch and I'd be happy to update the tests to always expect the span. I understand its a bit of a performance hit with adding potentially totally unneeded elements to the dom, but I'm not sure there is a better way to sidestep the google translate issue. It looks like a bug has been opened with Chrome (https://bugs.chromium.org/p/chromium/issues/detail?id=872770) - but until that's merged and released and the majority of the internet has updated, I think we should probably swap to "always use spans". We can keep a revert of this branch ready to roll in a few months when that is merged and out. Let me know, and thanks again!

wooorm:
This sounds like a problem with react, I don’t think we should merge this.

PR #372, opened 2019-12-10
Update type definitions with inline documentation

lsanwick:
I've simply used the documentation on the README and inserted it into the type definitions so that tools like VSCode will pick it up and display it inline when using the component.


wooorm:
Hi, thanks for your patience!

PR #352, opened 2019-10-13
Add typescript definition for html-parser

peng1999:
Fixes #340 .

PR #378, opened 2020-01-03
Fix typescript declaration

### LINKED ISSUE #377: Typescript problems
Getting this with React-Markdown 4.3.0 and Typescript 3.7.4.

Not sure what additional detail I could help with here :/

```
ERROR in node_modules/react-markdown/index.d.ts
ERROR in node_modules/react-markdown/index.d.ts(138,16):
TS2665: Invalid module name in augmentation. Module 'react-markdown/with-html' resolves to an untyped module at 'node_modules/react-markdown/with-html.js', which cannot be augmented.

ERROR in node_modules/react-markdown/index.d.ts
ERROR in node_modules/react-markdown/index.d.ts(140,35):
TS2307: Cannot find module 'react-markdown'.
```

peng1999:
As @petersowa pointed out in https://github.com/rexxars/react-markdown/pull/352#pullrequestreview-337828385 , I made a mistake in #352 . This should fix the bug.

Fixes #377

### DIFF
--- a/with-html.d.ts
+++ b/with-html.d.ts
@@ -1,0 +1,5 @@
+import {Component} from 'react'
+import ReactMarkdownRoot from 'react-markdown'
+
+export default class ReactMarkdown extends Component<ReactMarkdownRoot.ReactMarkdownProps, {}> {}
+
--- a/with-html.d.ts
+++ b/with-html.d.ts
@@ -1,5 +1,5 @@
 import {Component} from 'react'
-import ReactMarkdownRoot from 'react-markdown'
+import ReactMarkdownRoot = require('react-markdown')
 
-export default class ReactMarkdown extends Component<ReactMarkdownRoot.ReactMarkdownProps, {}> {}
-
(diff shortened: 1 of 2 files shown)


with-html.d.ts
```diff
@@ -0,0 +1,5 @@
+import {Component} from 'react'
+import ReactMarkdownRoot from 'react-markdown'
+
+export default class ReactMarkdown extends Component<ReactMarkdownRoot.ReactMarkdownProps, {}> {}
```
AviVahl: the actual `with-html.js` file has commonjs export, so this line should probably be:
```ts
declare class ReactMarkdown extends Component<ReactMarkdownRoot.ReactMarkdownProps, {}> {};
export = ReactMarkdown;
```
could/should probably also change the import to:
```ts
import ReactMarkdownRoot = require('react-markdown');
```

This way, the `.d.ts` will be compatible with `esModuleInterop` on or off.

rexxars: Thanks for the review @AviVahl :)

PR #386, opened 2020-01-24
allow nesting tags and close correct tags when multiple tags exist

### LINKED ISSUE #302: HTML parser: certain nested HTML elements cause fatal crash
Nesting certain HTML elements within the source string that is sent to ReactMarkdown causes a fatal crash. In particular I have noticed issues with the mark and span tags, though even divs can cause this crash in some circumstances.

This occurs when passing a string that includes HTML tags as the `source` prop to the ReactMarkdown element, with `escapeHtml` set to `false`. The project in which I encountered this issue is using ReactMarkdown 4.0.6, though the same behavior is replicable in the [live editor](https://rexxars.github.io/react-markdown/), replacing the example blockquote with other HTML.

Here are some examples of HTML that causes ReactMarkdown to crash:
- `<span><span>Text</span></span>`
- `<div><div>Text</div></div>`
- `<mark><mark>Text</mark></mark>`

divs seem to usually avoid the crash if there is whitespace between the tags. The following all render properly:
- `<div> <div> Text </div> </div>`
- `<div> <div>Text</div> </div>`
- `<div><div>Text</div> </div>`

Although this crashes:
- `<div><div> Text </div></div>`

spans, marks, and other inline elements crash even with spacing. The following all crash:
- `<mark> <mark> Text </mark> </mark>`
- `<span> <span> Text </span> </span>`
- `<em> <em> Text </em> </em>`

Here is the error that my project produces in the console when this crash occurs:

> ast-to-react.js:214 Uncaught TypeError: Cannot read property 'props' of undefined
>     at mergeNodeChildren (ast-to-react.js:214)
>     at getNodeProps (ast-to-react.js:175)
>     at astToReact (ast-to-react.js:18)
>     at ast-to-react.js:23
>     at Array.map (<anonymous>)
>     at resolveChildren (ast-to-react.js:22)
>     at astToReact (ast-to-react.js:19)
>     at ast-to-react.js:23
>     at Array.map (<anonymous>)
>     at resolveChildren (ast-to-react.js:22)

fabiosantoscode:
This allows for more liberal nesting of tags and markdown, such as:

* `<span><span>text</span></span>` (which crashed)
* `<div>[some markdown paragraphs with a HTML <span> in them</div>` (which didn't function properly)

I've noticed this project is supposed to be completely ES5, but I've used a Map to keep a record of closing tags at each level. This can be fixed in other ways, like using a Map ponyfill.

Fixes #302

### DIFF
--- a/src/plugins/html-parser.js
+++ b/src/plugins/html-parser.js
@@ -28,6 +28,7 @@ const defaultConfig = {
 }
 
 function parseHtml(config, tree, props) {
+  const openersByParent = new Map()
   let open
   let currentParent
   visit(
@@ -49,7 +50,12 @@ function parseHtml(config, tree, props) {
       }
 
       if (currentParent !== parent) {
-        open = []
+        if (openersByParent.has(parent)) {
+          open = openersByParent.get(parent)
+        } else {
+          open = []
+          openersByParent.set(parent, open)
+        }
         currentParent = parent
       }
 
@@ -68,7 +74,7 @@ function parseHtml(config, tree, props) {
         return true
       }
 
-      const matching = findAndPull(open, current.tag)
+      const matching = current.opening && findAndPull(open, current.tag)
 
       if (matching) {
         parent.children.splice(index, 0, parsedHtml(current, matching, parent))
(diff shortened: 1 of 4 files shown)


jorgeorpinel:
Hi. This seems old. But no conflicts so should we merge it?

fabiosantoscode:
Only hitch I see here is the fact that I used a Map. I could've also used an object if I created an ID property in every tag, but it seemed wasteful.

jorgeorpinel:
I didn't realize this isn't our repo until now 😆

PR #350, opened 2019-10-05
Add rel tag to links

juancarlosqr:
Hi,

For security reasons external links should be able to have a `rel` tag (.e.g. "noopener noreferrer"). More info [here](https://mathiasbynens.github.io/rel-noopener/)

I created a `linkRel` option to add the tag, similar to `linkTarget`

This is the first year I participate in the Hacktoberfest, so I'm open to any comments or feedback regarding my contribution 😉

Happy Hacktoberfest!

PR #407, opened 2020-03-31
Set rel attribute when linkTarget is specified

ibrahima:
This closes a potential security vulnerability. See the below for references.

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md
https://mathiasbynens.github.io/rel-noopener/

Note that this isn't checking whether linkTarget returns "_blank" in the case that it's a function. Might be worthwhile to check though.


mxmtsk:
You might want to take a look at #350 which adds support for what you want to do.

wooorm:
Same answer as what this is a duplicate of: https://github.com/remarkjs/react-markdown/pull/350#issuecomment-705571004

PR #417, opened 2020-04-26
Fix `renderers` prop in index.d.ts

dhruvio:
Updated it to user the `Renderers` interface, which allows specifying a renderer as a string, which is useful for conditionally overriding a renderer. For example:

```tsx
const renderers: Renderers = { 
  image: shouldRedactImages // boolean
      ? () => {
          return (<p className='text-secondary font-weight-bold'>[Image Redacted]</p>);
        }   
      : ReactMarkdown.renderers.image //Currently fails to typecheck because the default is a string.
};
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -88,7 +88,7 @@ declare namespace ReactMarkdown {
     readonly transformLinkUri?: ((uri: string, children?: ReactNode, title?: string) => string) | null
     readonly transformImageUri?: ((uri: string, children?: ReactNode, title?: string, alt?: string) => string) | null
     readonly unwrapDisallowed?: boolean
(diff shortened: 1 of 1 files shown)


kiliancs:
Duplicate of https://github.com/rexxars/react-markdown/pull/417

dhruvio:
@kiliancs I can see you marked this as a duplicate of #341. It seems like #341 doesn't use the `Renderers` type. IMO, it would be more DRY to use the `Renderers` type for this prop.

kiliancs:
I only suggested there were 3 PRs with the same problem and 2 fixes that should be consolidated. It seems like now there's only 2 of the PR left. I also preger the `Renderers` solution.

wooorm:
Hi, thanks for your patience!

Issue #427, opened 2020-05-17
Idea: Leverage rehype to simplify transforms


ChristianMurphy:
Hey @rexxars :wave: 
While looking through the project and saw an area to open up some new use cases and simplify the project in the process.

Currently the flow through `react-markdown` is

```
           +----------+        +----------------+        +--------------------------------------+
           |          |        |                |        |                                      |
-markdown->+  remark  +-mdast->+ remark plugins +-mdast->+ custom to html and to create element +-react element->
           |          |        |                |        |                                      |
           +----------+        +----------------+        +--------------------------------------+
```

where the translation to from markdown to html, the sanitizing, and the rendering to react elements all happen in the custom code section.

An alternative could be

```
           +----------+        +----------------+        +---------------+       +----------------+       +--------------+
           |          |        |                |        |               |       |                |       |              |
-markdown->+  remark  +-mdast->+ remark plugins +-mdast->+ remark-rehype +-hast->+ rehype plugins +-hast->+ rehype-react +-react elements->
           |          |        |                |        |               |       |                |       |              |
           +----------+        +----------------+        +---------------+       +----------------+       +--------------+
```

breaking out the translation from markdown to html into a step, and leveraging https://github.com/remarkjs/remark-rehype
allowing rehype plugins to further adapt the html (useful for sanitizing and extensions)
then translating the html to react elements https://github.com/rehypejs/rehype-react

The desired flow, markdown to react elements, with customization points is preserved.
The translation can be offloaded to rehype, reducing the amount and complexity of the code.
And it allows for customization, like [`remark-math`](https://github.com/remarkjs/remark-math) which need to do transforms both on the markdown and html.

I built version of what this could look like here: https://github.com/ChristianMurphy/react-remark (with live demos https://christianmurphy.github.io/react-remark)
If you'd be interested I could open a pull request to pull the code over here.

Thoughts? :thought_balloon:

rexxars:
Hi @ChristianMurphy !

Thanks for taking the time to make such a thorough issue (and pull request!).

You seem to be quite familiar with the unified ecosystem and already have made this into a separate module (`react-remark`), and I was wondering what your thoughts are on remark-react, react-remark and now react-markdown?

[...]

PR #452, opened 2020-07-30
Update html-parser.js: add preprocessingInstructions config of `html-to-react`

fengma1992:
add preprocessingInstructions config of `html-to-react`


ChristianMurphy (commented):
Thanks for your patience @fengma1992!
Could you expand on what you are trying to solve here?
Are you looking for a way to do transforms on HTML? Something like #427 ?

Issue #327, opened 2019-07-31
Mis-matched renderers types


mykter:
I'm having problems with the types of my custom renderers, but I did this sanity check using the default renderers which suggests to me the types are probably inconsistent.

In `index.d.ts` we have:
```ts
export interface ReactMarkdownProps {
    ...
    readonly renderers?: {[nodeType: string]: ReactType}
}
...
interface Renderers {
    [key: string]: string | Renderer<any>
  }
...
export var renderers: Renderers
```

Which leads to a problem when you try and do:
```tsx
<ReactMarkdown source={source} renderers={ReactMarkdown.renderers} />
```
```
Type 'Renderers' is not assignable to type '{ [nodeType: string]: ElementType<any>; }'.
  Index signatures are incompatible.
    Type 'string | Renderer<any>' is not assignable to type 'ElementType<any>'.
      Type 'string' is not assignable to type 'ElementType<any>'.ts(2322)
index.d.ts(82, 14): The expected type comes from property 'renderers' which is declared here on type 'IntrinsicAttributes & IntrinsicClassAttributes<ReactMarkdown> & Readonly<ReactMarkdownProps> & Readonly<{ children?: ReactNode; }>'
```

It might be that this pull request fixes it: https://github.com/rexxars/react-markdown/pull/261

ChristianMurphy:
This may also be resolved by #428, which converts the project to native TypeScript.

n-lam:
Since #428 was closed, I was wondering if we could reopen #261 as a temporary solution while waiting for the complete refactor? I'm also happy to open a new PR to make additional changes if needed.

ChristianMurphy:
> Since #428 was closed

https://github.com/remarkjs/react-markdown/pull/428#issuecomment-717633714 the work is still planned, the closing was a quirk of GitHub and branch renaming.

wooorm:
this should be solved by https://github.com/remarkjs/react-markdown/pull/566

PR #428, opened 2020-05-17
Typescript and Rehype refactor

ChristianMurphy:
Unified, through remark and rehype is providing the translation of markdown source into react elements.
TypeScript is used to allow for catching type related bugs and automatic generation of typings files.
tsdx is mananging the build process.
Storybook is leveraged to create runnable examples which can be hosted on GitHub pages.
GitHub actions are used for continuous integration and automatic deployment of the demo site.

There is a live demo of this code running at https://christianmurphy.github.io/react-remark
and a scoped version available on the registry with `npm install --save @christianmurphy/react-remark`

resolves #427

PR #461, opened 2020-08-29
Update TypeScript types for Renderers type

### LINKED ISSUE #439: Improve TS type for ReactMarkdown.Renderers
I've noticed the following implementation for 

```ts
interface Renderers {
    [key: string]: string | Renderer<any>
  }
```

Know that you can use the following instead:

```ts
interface Renderers {
  [key in ReactMarkdown.NodeType]?: string | ReactMarkdown.Renderer<any>;
}
```

This has the advantage of allowing intellisense for `Renderers` (e.g: `Renderers.link`, etc. Basically, keys can be autocompleted.

JacobFischer:
This should resolve #439 and #327.

I see that this PR basically steps on the toes of #341,  #355, and #417. However none of those seemed to actually type all the possible tag specific props. E.g. `level` in `heading`.

Instead, I've gone through each tag and examined the props being sent to them, and then documented those possible props with their types in these enhanced typings. I believe I've got them all documented, but if I've missed anything feel free to request changes.

As these changes are 100% TypeScript typings related they should have no impact on runtime.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -7,7 +7,7 @@
 // - Espen Hovlandsdal <https://espen.codes/>
 // - Ted Piotrowski <https://github.com/ted-piotrowski>
 
-import {Component, ReactElement, ReactNode, ReactType} from 'react'
+import { Component, ElementType, ReactNode } from 'react'
 
 declare class ReactMarkdown extends Component<ReactMarkdown.ReactMarkdownProps, {}> {}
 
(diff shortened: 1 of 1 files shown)


wooorm:
Have you seen https://github.com/remarkjs/react-markdown/pull/428/files?

wooorm:
Hi, thanks for your patience!

JacobFischer:
That's fine. I'm cloning down that fork you mentioned now to probe into what it does/changes to see if it includes the types I added here. If it does not I might try to make another PR there to fix their types.

This PR was only concerned with fixing types with as few changes as possible (actually none at run/build time). That PR is a full re-write so they are quite different in scope.

Issue #477, opened 2020-10-15
Move generated demo files from source


wooorm:
### Subject of the feature

Currently, the demo code is checked in. I propose moving the demo to an orphan branch, and automatically publishing in a cron job on travis.

### Problem

Noise in git diffs.

### Expected behavior

Orphan branch, build demo in it, automatically publish w/ travis.

### Alternatives

Actions or so but travis should be fine. Orphan branch also has worked well for me in the past.

PR #486, opened 2020-10-16
Move demo to `website` branch

wooorm:
* Also move UMD bundle to root to remove empty folder

Closes GH-477.


ChristianMurphy:
How will `website` branch differ from `gh-pages` branch? Or will they be the same?

wooorm:
Website is clean: no checked in built files. gh-pages is dirty: because GH requires served files to be checked in

wooorm:
(this is already live btw, so the difference can be seen: https://github.com/remarkjs/react-markdown/tree/gh-pages vs https://github.com/remarkjs/react-markdown/tree/website)

ChristianMurphy:
So there will be three branches?
`main` for the library source code, `website` for the website source code, and `gh-pages` for the website deployment?

wooorm:
Yeah, that’s my plan! You have to have gh-pages, because the generated files are there. Moving the demo files to a different branch separates the two things: one a library, another a (small) site, which have very different goals

ChristianMurphy (approved):
fine by me.
I've usually put the site source in a folder on the `main` branch, but I'd be open to trying a dedicated branch for the site source.

PR #490, opened 2020-10-17
test(types): add dtslint tests to verify typescript typings

### LINKED ISSUE #482: Add type tests, `types/` folder
### Subject of the feature

Types are not checked at all. They’re in the root.

### Problem

Clean, test

ChristianMurphy:
resolves #482

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -43,7 +43,6 @@ declare namespace ReactMarkdown {
       | ((uri: string, children?: ReactNode, title?: string, alt?: string) => string)
       | null
     readonly renderers?: {[nodeType: string]: ElementType}
-    /** @deprecated use plugins */
     readonly astPlugins?: PluggableList
     readonly plugins?: PluggableList
   }
@@ -66,10 +65,13 @@ declare namespace ReactMarkdown {
     readonly unwrapDisallowed?: boolean
   }
 
+  interface AllowDangerousHtmlProp {
+    readonly allowDangerousHtml?: boolean
+  }
+
   interface EscapeHtmlProp {
     /** @deprecated use allowDangerousHtml */
     readonly escapeHtml?: boolean
-    readonly allowDangerousHtml?: boolean
   }
 
   interface SkipHtmlProp {
@@ -79,7 +81,7 @@ declare namespace ReactMarkdown {
   type ReactMarkdownProps = ReactMarkdownPropsBase &
     MutuallyExclusive<ChildrenProp, SourceProp> &
     MutuallyExclusive<AllowedTypesProp, DisallowedTypesProp> &
-    MutuallyExclusive<EscapeHtmlProp, SkipHtmlProp>
+    MutuallyExclusive<EscapeHtmlProp, MutuallyExclusive<SkipHtmlProp, AllowDangerousHtmlProp>>
 
   const types: NodeType[]
   const renderers: Renderers
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,11 +5,13 @@ import {PluggableList} from 'unified'
 import * as unist from 'unist'
 import * as mdast from 'mdast'
 
-declare namespace ReactMarkdown {
-  type Not<T> = {
-    [key in keyof T]?: never
-  }
+type Not<T> = {
+  [key in keyof T]?: never
+}
 
+type MutuallyExclusive<T, U> = (T & Not<U>) | (U & Not<T>)
+
+declare namespace ReactMarkdown {
   type Point = unist.Point
 
   type Position = unist.Position
@@ -75,12 +77,9 @@ declare namespace ReactMarkdown {
   }
 
   type ReactMarkdownProps = ReactMarkdownPropsBase &
-    ((SourceProp & Not<ChildrenProp>) | (ChildrenProp & Not<SourceProp>)) &
-    (
-      | (AllowedTypesProp & Not<DisallowedTypesProp>)
-      | (DisallowedTypesProp & Not<AllowedTypesProp>)
-    ) &
-    ((EscapeHtmlProp & Not<SkipHtmlProp>) | (SkipHtmlProp & Not<EscapeHtmlProp>))
+    MutuallyExclusive<ChildrenProp, SourceProp> &
+    MutuallyExclusive<AllowedTypesProp, DisallowedTypesProp> &
+    MutuallyExclusive<EscapeHtmlProp, SkipHtmlProp>
 
   const types: NodeType[]
   const renderers: Renderers
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,16 +1,8 @@
-// Type definitions for react-markdown > v3.3.0
-// Project: https://github.com/remarkjs/react-markdown
-// Definitions by:
-// - Ruslan Ibragimov <https://github.com/IRus>
-// - Kohei Asai <me@axross.io>
-// - ClassicDarkChocolate <https://github.com/ClassicDarkChocolate>
-// - Espen Hovlandsdal <https://espen.codes/>
-// - Ted Piotrowski <https://github.com/ted-piotrowski>
-// - Christian Murphy <https://github.com/ChristianMurphy>
+// TypeScript Version: 3.4
 
-import {Component, ReactElement, ReactNode} from 'react'
+import {Component, ElementType, ReactNode} from 'react'
 import {PluggableList} from 'unified'
-import {Content} from 'mdast'
+import * as mdast from 'mdast'
 
 declare namespace ReactMarkdown {
   interface Point {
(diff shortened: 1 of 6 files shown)


react-markdown-types-test.tsx
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react'
+import * as ReactDom from 'react-dom'
+import * as ReactMarkdown from 'react-markdown'
+import * as ReactMarkdownWitHtml from 'react-markdown/with-html'
+
+ReactDom.render(<ReactMarkdown># header</ReactMarkdown>, document.getElementById('main'))
+ReactDom.render(<ReactMarkdownWitHtml># header</ReactMarkdownWitHtml>, document.getElementById('main'))
```
wooorm, line 7: will `document.body` or so also be fine for the tests? I like it because it’s a bit shorter and (while it’s all theoretically) would always work

package.json
```diff
@@ -61,17 +61,20 @@
...
     "babel-plugin-transform-react-remove-prop-types": "^0.4.24",
+    "dtslint": "^4.0.4",
     "enzyme": "^3.10.0",
     "enzyme-adapter-react-16": "^1.14.0",
     "eslint": "^7.0.0",
     "eslint-config-prettier": "^6.0.0",
     "eslint-config-sanity": "^4.0.2",
     "eslint-plugin-react": "^7.14.2",
     "jest": "^26.0.1",
+    "npm-run-all": "^4.1.5",
```
wooorm, line 75:
```suggestion
    "npm-run-all": "^4.0.0",
```

package.json
```diff
@@ -84,6 +87,7 @@
     "rollup": "^2.0.0",
     "rollup-plugin-node-polyfills": "^0.2.0",
     "rollup-plugin-terser": "^7.0.0",
+    "typescript": "^4.0.3",
```
wooorm, line 90:
```suggestion
    "typescript": "^4.0.0",
```

wooorm: is this even needed? doesn’t dtslint try with lots of different versions?

package.json
```diff
@@ -61,17 +61,20 @@
     "@rollup/plugin-json": "^4.0.0",
     "@rollup/plugin-node-resolve": "^9.0.0",
     "@rollup/plugin-replace": "^2.0.0",
+    "@types/react": "^16.9.53",
+    "@types/react-dom": "^16.9.8",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.24",
+    "dtslint": "^4.0.4",
```
wooorm, line 67:
```suggestion
    "dtslint": "^4.0.0",
```

package.json
```diff
@@ -61,17 +61,20 @@
     "@rollup/plugin-json": "^4.0.0",
     "@rollup/plugin-node-resolve": "^9.0.0",
     "@rollup/plugin-replace": "^2.0.0",
+    "@types/react": "^16.9.53",
+    "@types/react-dom": "^16.9.8",
```
wooorm, line 65: I guess these are kinda like peer deps, right? In that case, I’m assuming using the `16.0.0` versions will reduce noice in the future

index.d.ts
```diff
@@ -25,20 +17,20 @@ declare namespace ReactMarkdown {
     readonly indent?: number[]
   }
 
-  export type NodeType = Content['type']
+  type NodeType = Content['type']
 
-  export type AlignType = 'left' | 'right' | 'center' | null
+  type AlignType = 'left' | 'right' | 'center' | null
```
wooorm, line 22: Can these be imported (and `ReferecenType`)? https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f37828cb0a0d60300a749c2d3f12d9d4314a59c2/types/mdast/index.d.ts#L9

package.json
ChristianMurphy: It does try different versions,
but DTSLint made it a peerDependency in the latest version, and requires the peer to run, not entirely sure why :man_shrugging: 
https://github.com/microsoft/dtslint/blob/952086da5617f3f2c94bc9aad849fe51732eb00e/package.json#L40-L42

wooorm: oh, interesting. So we’ll hit this with other unified stuff too, good to know!

package.json
ChristianMurphy: Added as peer dependencies, for testing I'd recommend staying with React 16.
It will reduce noise and hopefully ensure a smooth transition to react 17 ([coming soon](https://reactjs.org/blog/2020/08/10/react-v17-rc.html))

package.json
```diff
@@ -49,7 +49,8 @@
     "xtend": "^4.0.1"
   },
   "peerDependencies": {
-    "react": "^15.0.0 || ^16.0.0"
+    "react": "^15.0.0 || ^16.0.0",
+    "@types/react": "^15.0.0 || ^16.0.0"
```
wooorm, line 53: btw, isn’t 17 around the corner? Could we allow that? (e.g., `>= 15.0.0` or so)?

ChristianMurphy: It is, see https://github.com/remarkjs/react-markdown/pull/490#discussion_r506969473

We could add it :thinking:
I haven't tested it yet to ensure it still works.
May hope is it should be painless transition, that will only require an update to the peer dependencies.

wooorm: could be added later!

wooorm:
still blocked? looks good to me!

wooorm (approved):

ChristianMurphy:
@wooorm 4639733 was the other set of changes I wanted to get in, leveraging more unist types and producing error when properties that are mutually exclusive are used together.
This should be ready to go now, through more tests for other properties are definitely welcome!

index.d.ts
```diff
@@ -1,74 +1,92 @@
...
+    readonly skipHtml?: boolean
+  }
+
+  type ReactMarkdownProps = ReactMarkdownPropsBase &
+    ((SourceProp & Not<ChildrenProp>) | (ChildrenProp & Not<SourceProp>)) &
+    (
+      | (AllowedTypesProp & Not<DisallowedTypesProp>)
+      | (DisallowedTypesProp & Not<AllowedTypesProp>)
+    ) &
+    ((EscapeHtmlProp & Not<SkipHtmlProp>) | (SkipHtmlProp & Not<EscapeHtmlProp>))
```
ChristianMurphy, line 83: `Not` allows preventing the other type from being applied at the same time.
It's a bit verbose, but may be needed until Exact types https://github.com/microsoft/TypeScript/issues/12936 become a thing

index.d.ts
```diff
@@ -1,74 +1,92 @@
...
+
+  const types: NodeType[]
+  const renderers: Renderers
+  function uriTransformer(uri: string): string
 }
 
-declare class ReactMarkdown extends Component<ReactMarkdown.ReactMarkdownProps, {}> {}
+declare function ReactMarkdown(props: ReactMarkdown.ReactMarkdownProps): ReactElement
```
ChristianMurphy, line 90: this is roughly the equivalent of
```ts
declare const ReactMarkdown: React.FunctionComponent<ReactMarkdownProps>
```

However, `var`/`let`/`const` cannot have an overlapping name with `declare namespace`, while `function` does support it.

ChristianMurphy: the previous `declare class` is mismatches with the actual implementation
https://github.com/remarkjs/react-markdown/blob/4f5dbe21d0375f08e44f453e41505f2b45c1230e/src/react-markdown.js#L19

index.d.ts
ChristianMurphy: cleaned this up with a `MutuallyExclusive` type in eead2f3

index.d.ts
```diff
@@ -1,74 +1,91 @@
...
+  }
+
+  interface SkipHtmlProp {
+    readonly skipHtml?: boolean
+  }
+
+  type ReactMarkdownProps = ReactMarkdownPropsBase &
+    MutuallyExclusive<ChildrenProp, SourceProp> &
+    MutuallyExclusive<AllowedTypesProp, DisallowedTypesProp> &
+    MutuallyExclusive<EscapeHtmlProp, SkipHtmlProp>
```
ChristianMurphy, line 82: Prevents exclusive types from overlapping.
It's a bit verbose, but may be needed until Exact types microsoft/TypeScript#12936 become a thing

index.d.ts
```diff
@@ -30,35 +27,62 @@ declare namespace ReactMarkdown {
...
     readonly transformLinkUri?:
       | ((uri: string, children?: ReactNode, title?: string) => string)
       | null
     readonly transformImageUri?:
       | ((uri: string, children?: ReactNode, title?: string, alt?: string) => string)
       | null
-    readonly unwrapDisallowed?: boolean
     readonly renderers?: {[nodeType: string]: ElementType}
+    /**
+     * @deprecated please use plugins
```
wooorm: `astPlugins` was never documented in the props, but it is still recommended by the readme in the HTML section as that’s the only way to have HTML in Markdown now.

There are a lot of bugs with it though.

Also: I soft deprecated `escapeHtml` in favor of `allowDangerousHtml`

ChristianMurphy: resolved in 2872089

wooorm (approved):

Issue #495, opened 2020-10-22
The allowedTypes and unwrapDisallowed props cannot be used together in TypeScript


js-williams:
### Subject of the issue

The typings for the props for `ReactMarkdown` seem to mismatch with the actual code. They don't allow `allowedTypes` and `unwrapDisallowed` to be used together, but this seems to be okay in the source code.

Code references:
* `allowedTypes` and `disallowedTypes` cannot be used together - https://github.com/remarkjs/react-markdown/blob/main/src/react-markdown.js#L23-L25

* However, it should be fine to use the `unwrapDisallowed` prop with the `allowedTypes` prop - https://github.com/remarkjs/react-markdown/blob/main/src/react-markdown.js#L48-L58

### Your environment

*   **OS**: MacOS 10.14.5
*   **Packages**: react-markdown 5.0.0
*   **Env**: node 10.20.1 npm 6.12.0 TypeScript 4.0.3

### Steps to reproduce

The following TypeScript code will throw a type error when compiled:
```ts
<ReactMarkdown source="example" allowedTypes={['strong', 'text']} unwrapDisallowed={true} />
```

### Expected behavior

The compiler should allow allowedTypes and unwrapDisallowed to be used together.

### Actual behavior

A type error is thrown on compilation.

ChristianMurphy:
Testing at https://codesandbox.io/s/infallible-swanson-1hr2m it appears that `unwrapDisallowed` can take effect with `allowedTypes`.

If you're interested a PR would be welcome.
Move https://github.com/remarkjs/react-markdown/blob/e033f5669143c4d9134a69fdc9b625ad436428e4/index.d.ts#L65 into https://github.com/remarkjs/react-markdown/blob/e033f5669143c4d9134a69fdc9b625ad436428e4/index.d.ts#L32-L48
and update
https://github.com/remarkjs/react-markdown/blob/e033f5669143c4d9134a69fdc9b625ad436428e4/react-markdown-types-test.tsx#L31-L36

wooorm:
released in 5.0.2!

PR #496, opened 2020-10-22
Move disallowedTypes into base props in typings

js-williams:
Closes GH-495


ChristianMurphy (approved):
Thanks @js-williams! :bow:

wooorm:
thanks, released in 5.0.2!

Issue #497, opened 2020-10-24
Update React version to React 17 (peerDependencies)


Dev-CasperTheGhost:
### Subject of the feature

Update React version to React 17 in peerDependencies

### Problem

react-markdown does not support React 17

### Expected behavior

Support the latest version of React.

### Alternatives

/

wooorm:
Do you get a warning that React 17 isn’t supported, or does something actually crash?

Dev-CasperTheGhost:
Hi,

I'm getting an error when installing it from npm (see image).

Thanks for your fast response.

![image](https://user-images.githubusercontent.com/53900565/97081812-2ff27b00-1605-11eb-9f71-0b7114f8f8b7.png)

wooorm:
Are you on npm 7 already? (`npm --version`)

If so, try the `--legacy-peer-deps` for now.

---

I was looking into React 17 a day or so ago, and it all worked well. However, this project is using `enzyme-adapter-react-16` and in dev-dependencies, and there isn’t a `enzyme-adapter-react-17` yet, so I was waiting on that.

Dev-CasperTheGhost:
Alright, will do. Thanks for your help.

wooorm:
Released!

Dev-CasperTheGhost:
Thankyou!

Issue #313, opened 2019-06-03
HTML parser: incompatible w/ React 15


ujovlado:
Hi, looks like react-markdown is not compatible with React 15, because html-to-react 1.3.4. requires React 16 as peerDependency.

https://github.com/aknuds1/html-to-react/blob/eea2ee18196bb002746bdcb91a8a596e4a8cdd61/package.json#L44

ChristianMurphy:
#428 may resolve this, as it removes the dependency on html-to-react

ChristianMurphy:
With React 16 being stable, and [React 17 around the corner](https://reactjs.org/blog/2020/08/10/react-v17-rc.html), I'm not sure if officially trying to support React <16 makes sense.
Maintaining support for a version of React that is no longer supported by the React team itself, seems outside the scope of this project.

wooorm:
The suggestion here is to upgrade React. React 17 is indeed around the corner, React 16 is in a good place, and React 15 is from 4 years ago.
I am not too happy with the fact that `html-to-react` has this peer dependency, the plan is to change that to a rehype version in the next major, but it’s not something we can change without breaking things.
If you really need to, you could use an older unmaintained version of react-markdown. Or: don’t support dangerous HTML.

ujovlado:
The issue wasn't about support for React 15. It was about "claim" that specific version is compatible with React 15, but it wasn't. 😉 

In the meantime, we upgraded to latest React, so everything works as expected (no warnings, etc.).

Thanks for the great library!

PR #498, opened 2020-10-24
Add tests on React 17

rexxars:
Upgrades the peer dependency version range to support React 17, and uses React 17 for the tests. Switches to using the unofficial `@wojtekmaj/enzyme-adapter-react-17` adapter for enzyme, for now (until the official adapter is released).

On a slightly unrelated note - I'm not sure I agree that `@types/react` should be a peer dependency?

Closes GH-497.


rexxars:
Hmm, this seems to be failing with npm 7 because of upstream peer dependencies, too :/

package.json
```diff
@@ -64,19 +64,19 @@
     "@rollup/plugin-replace": "^2.0.0",
     "@types/react": "^16.0.0",
     "@types/react-dom": "^16.0.0",
+    "@wojtekmaj/enzyme-adapter-react-17": "^0.1.1",
```
ChristianMurphy, line 67: it sounds like the upstream version is in progress https://github.com/enzymejs/enzyme/pull/2430

ChristianMurphy (approved):

package.json
```diff
@@ -50,7 +50,7 @@
   },
   "peerDependencies": {
     "@types/react": "^15.0.0 || ^16.0.0",
-    "react": "^15.0.0 || ^16.0.0"
+    "react": "^15.0.0 || ^16.0.0 || ^17.0.0"
```
wooorm, line 53: I believe 15 isn‘t working, so might as well update that: https://github.com/remarkjs/react-markdown/issues/313

Can we switch to `>= 16`, so that we’re not bugged by React 18 either? (If 18 breaks, I’m sure we’ll get issues about that and can fix it)

package.json
wooorm: From reading some more comments there, it could take longer

package.json
```diff
@@ -64,21 +64,21 @@
...
     "jest": "^26.0.0",
     "npm-run-all": "^4.0.0",
     "prettier": "^2.0.0",
-    "react": "^16.0.0",
-    "react-dom": "^16.0.0",
     "react-katex": "^2.0.0",
-    "react-test-renderer": "^16.0.0",
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1",
+    "react-test-renderer": "^17.0.1",
```
ChristianMurphy, line 81:
```suggestion
    "react": "^17.0.0",
    "react-dom": "^17.0.0",
    "react-test-renderer": "^17.0.0",
```

package.json
```diff
@@ -39,18 +39,18 @@
   "dependencies": {
     "@types/mdast": "^3.0.3",
     "@types/unist": "^2.0.3",
-    "html-to-react": "^1.3.4",
+    "html-to-react": "^1.4.5",
     "mdast-add-list-metadata": "1.0.1",
     "prop-types": "^15.7.2",
-    "react-is": "^16.8.6",
+    "react-is": "^17.0.1",
```
ChristianMurphy, line 45:
```suggestion
    "react-is": "^17.0.0",
```

ChristianMurphy (approved):

wooorm (approved):
@rexxars Good to go, I think?

package.json
wooorm: @rexxars Oh: I do agree on these suggestions though!

rexxars:
> @rexxars Good to go, I think?

Yep. The react-katex dependency causes some issues with npm7, but only in development mode. Will create a separate PR to get rid of it and simplify the tests it's used in.

package.json
rexxars: Why? Aesthetics?

ChristianMurphy: Broadening the version range as much as possible, gives package managers more opportunity to de-duplicate dependencies.
Which in turn can help bundlers like CRA which will error it detects multiple react versions (without defining their own renderer).

rexxars: I see the point for dependencies, not sure I see the big deal with dev dependencies. Also, following that logic, shouldn't the version range be >=16, then? That's broader, and strictly speaking compatible?

ChristianMurphy: I personally avoid `>=` as much as possible preferring a combination of minor ranges `^16 || ^17`.
But I know @wooorm has a different opinion on that.

rexxars: 🤷  Not a hill I'm willing to die on 😉 
^17.0.0 works for me.

wooorm: Removing meaningless stuff in semver ranges helps clean up package locks, and indeed should help with deduping. And it’s just cleaner: there is really no need to be so specific, having the zeros there makes the major more pronounced, and it’s more clear for for cases that do need it (`"html-to-react": "^1.4.5"`).

For the difference between how the peer dep doesn’t have an upper boundary but dev-deps do: My experience it’s similar to `engines`: I can guarantee that it works on v16, and it will probably work afterwards as well. For example: we’re currently blocking folks from updating because there is an error, even though it works fine and react 17 didn‘t break us. and react 18 also probably won’t.

rexxars:
Not sure there's much we can do about the tests failing - seems to be because of upstream enzyme adapter peer dependency ranges. Did you want more changes here, @wooorm? Or should we leave it until the official enzyme adapter is out?

wooorm:
An official version seems to get closers: https://github.com/enzymejs/enzyme/pull/2430#issuecomment-716327508.

I’m fine with the changes here and using the other package now, but I’m unhappy with the failing tests 😬 The enzyme PR also mentions a PR to proptypes too, which seems to be coming after the enzyme update.
Can we change the [default install in Travis](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#dependency-management) to include the `--legacy-peer-deps` mention in the npm failure in Travis?

Hypnosphi:
@wooorm @rexxars can you please consider just updating peerDependencies range for now?

package.json
```diff
@@ -39,18 +39,18 @@
...
     "prop-types": "^15.7.2",
-    "react-is": "^16.8.6",
+    "react-is": "^17.0.0",
     "remark-parse": "^9.0.0",
     "unified": "^9.0.0",
     "unist-util-visit": "^2.0.0",
     "xtend": "^4.0.1"
   },
   "peerDependencies": {
     "@types/react": "^15.0.0 || ^16.0.0",
```
wooorm, line 52: /cc @rexxarsm, @ChristianMurphy I guess types should be unlocked too?

Btw: I’m going to unlock right now to at least hide the warnings. Tests seem to indicate it’s all fine, but then this PR can land later when enzyme updates

wooorm:
I’ve unlocked the package already, this PR can add tests for it later when enzyme updates.

https://github.com/remarkjs/react-markdown/commit/bb0bdded9f5ee43618e230d8043b6293f9e9eeb1

ChristianMurphy:
A possible way to unblock this would be to migrate from enzyme to testing library react, which supports React 17.
https://testing-library.com/docs/react-testing-library/intro/

ChristianMurphy:
@rexxars #519 should unblock the `enzyme` peer dependency issue, when you have a chance to rebase off `main`, this should be good to go.

ChristianMurphy:
Synced with develop, the `react-katex` test is now causing issues, opened #521 to address this

ChristianMurphy (approved):

Issue #100, opened 2017-11-16
How does the styling work?


JikkuJose:
There seems to be no mention about it? The demo code seems to be too cryptic!

rexxars:
There isn't any styling by default. react-markdown just renders semantic markup, then it's up to the developer to apply styles to those. I'll flag this for clarity in v3's readme.

JikkuJose:
Thanks, just wondering about an idea. How complicated is it to add classes to the rendered elements using props like so:

```
tagClasseNames={{
  h1: 'custom-head',
  p: 'custom-p'
}}
```

rexxars:
That would work, but every custom renderer would have to respect the class names. It also adds to an API that I'm hoping to keep fairly small. If you really need this, you could always just override the default renderers.

stahlmanDesign:
I installed this module with npm https://github.com/sindresorhus/github-markdown-css
and I import it like this
`import 'github-markdown-css'` <-- it finds the css in node_modules

Then wrap your markdown with the className 'markdown-body' used by the css
```jsx
<div className='markdown-body'>
  <ReactMarkdown source='# hello'/>
</div>
```
This styles the `th` `td` `table` etc like github markdown

Taskkill:
> There isn't any styling by default. react-markdown just renders semantic markup, then it's up to the developer to apply styles to those. I'll flag this for clarity in v3's readme.

Can you please put it in the README? Idk if it was in previous versions, but it is not in current readme.
Honestly confused me a lot at first, I'll admit I do not have much knowledge about MarkDown so it didn't hit me immediately that it renders only semantic MarkDown. Anyways thank you.

OlivierJM:
> I installed this module with npm https://github.com/sindresorhus/github-markdown-css
> and I import it like this
> `import 'github-markdown-css'` <-- it finds the css in node_modules
> 
> Then wrap your markdown with the className 'markdown-body' used by the css
> 
> ```js-jsx
> <div className='markdown-body'>
>   <ReactMarkdown source='# hello'/>
> </div>
> ```
> 
> This styles the `th` `td` `table` etc like github markdown

This fixed my issue, thanks a lot.

tnrich:
@rexxars 
If you could link to this issue and the fix @stahlmanDesign mentioned above in the README that would be awesome!

wooorm:
[...]

Issue #325, opened 2019-07-31
Editor in Demo included?


beamercola:
Is the editor in the demo included in the library?  It's great.

ChristianMurphy:
The editor included in the demo is https://github.com/codemirror/codemirror

PR #502, opened 2020-11-03
Add a small note that no css rules are included by default

tnrich:
@wooorm  @rexxars 

I was personally surprised when my markdown didn't look like the markdown in demo page (https://remarkjs.github.io/react-markdown/) and I had to dig through the issues to find out how to get my markdown to look nice. I think that is probably the case for many other people judging by how popular @stahlmanDesign's comment was (https://github.com/remarkjs/react-markdown/issues/100#issuecomment-504505123). 

Adding a small note to the readme would go a long way to making this less of an issue. 

Thanks!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -413,6 +413,9 @@ Overwriting `transformLinkUri` or `transformImageUri` to something insecure or
 turning `allowDangerousHtml` on, will open you up to XSS vectors.
 Furthermore, the `plugins` you use and `renderers` you write may be insecure.
 
+## Default Styling
+By default, react-markdown does not include any CSS. You can roll your own css solution or use something pre-built like https://github.com/remarkjs/react-markdown/issues/100#issuecomment-504505123
+
 ## Related
 
 *   [`MDX`](https://github.com/mdx-js/mdx)
--- a/readme.md
+++ b/readme.md
@@ -414,7 +414,8 @@ turning `allowDangerousHtml` on, will open you up to XSS vectors.
 Furthermore, the `plugins` you use and `renderers` you write may be insecure.
 
 ## Default Styling
-By default, react-markdown does not include any CSS. You can roll your own css solution or use something pre-built like https://github.com/remarkjs/react-markdown/issues/100#issuecomment-504505123
+
+By default, react-markdown does not include any CSS.  You can roll your own css solution or use something pre-built like https://github.com/remarkjs/react-markdown/issues/100#issuecomment-504505123
 
 ## Related
 


wooorm:
Appreciate you taking the time to work on a PR, Thomas!
But this seems to push through what you want, and not address my questions at all? (Ref: https://github.com/remarkjs/react-markdown/issues/100#issuecomment-721277840).

While opening your PR, you saw our contributing guides. If you dismissed them quickly, here’s a direct link: https://github.com/remarkjs/.github/blob/main/contributing.md#submitting-a-pull-request. Most notable I’m missing the last two points.

tnrich:
If you don't want the note then don't take it. I think it is helpful and I bet other people would think so too. Also I did take your feedback into consideration by explaining that no CSS is shipped by default and simply pointed to 1 possible solution. 

As for linting.. it's a readme change..?

wooorm:
> If you don't want the note then don't take it.

This is what I said before. You asked maintainers to add something, I explained why that particular solution doesn’t make sense to me.

> I think it is helpful and I bet other people would think so too.

Maybe, which is why I’m trying to help guide you to an actionable proposal.
I think this currently answers one of thousands of possible questions folks can have around react-markdown in one particular way. We can’t list all questions, or all answers. If we did, the readme wouldn’t be useful any more. I don’t think github’s CSS makes sense for everyone.

> Also I did take your feedback into consideration by explaining that no CSS is shipped by default and simply pointed to 1 possible solution.

I suggested that this one particular solution wasn’t what everyone else would want. You instead opted to push through and suggest that one solution.

> As for linting.. it's a readme change..?

Wait, so the docs aren’t important? Yeah, no.

tnrich:
You're welcome to come up with changes to it or fix it however you like. The original issue is from 2017 and I still had to dig through the issues to figure out if I was doing something wrong. I don't really care what the docs say so long as they have some little mention that the CSS in demo doesn't ship by default. 

Another good idea IMO is that the demo could also mention this. I think that would be helpful as well.

As for me trying to "push through what [I] want", you locked that issue otherwise I would have made a comment in that thread instead of trying to make a PR.

ChristianMurphy:
I agree with @wooorm,
A readme introduces what a project _is_.
it seems to be an exercise in futility to enumerate everything the project _is not_.

> I was personally surprised when my markdown didn't look like the markdown in demo page (https://remarkjs.github.io/react-markdown/)

The docs are (or at least intend to be) pretty up front about it being a renderer.
I'm not sure I follow the line of thinking that it is (or should be) a design library.
If you see a way to make it clearer what the project is, in the readme thoughts are welcome.

> I had to dig through the issues to find out how to get my markdown to look nice

Okay?
This library, like many libraries, has a specific purpose and assumes some prior knowledge.
This could be paraphrased as "How to write a stylesheet?", which is a valuable skill, but not one I think is a good fit for the readme.
I'd be open to the idea of a section on prerequisite knowledge, but even that seems to edge into explaining what the project _is not_.

> Adding a small note to the readme would go a long way to making this less of an issue

It isn't an issue :slightly_smiling_face: it's how the library works.

> I don't really care what the docs say so long as they have some little mention that the CSS in demo doesn't ship by default.

Again, this comes back to describing what the project isn't.
The editor in the readme isn't part of the project either (https://github.com/remarkjs/react-markdown/issues/325), yet that doesn't need to be a part of the readme.

---

I appreciate your thoughts @tnrich, and that you are investing in improving the project.
Our comments aren't meant to criticize, they are meant to help guide the documentation towards information that applies broadly, rather than too much on specific use cases (with 30K adopters and counting, there are a lot of use cases, too put in the readme and still have it readable).

tnrich:
Understood. That's a perfectly valid opinion to have. I still think it would probably be broadly useful and save people time but they can always go digging into the issues to find answers. Cheers :)

tnrich:
@ChristianMurphy can you link to the demo's source code?

ChristianMurphy:
https://github.com/remarkjs/react-markdown/tree/website

PR #507, opened 2020-11-14
Fix exception on missing element from `html-to-react`

cerkiewny:
I have noticed that if your html is incorrect - starts with close tag, the solution crushes,
I thought it might be better to just skip the problematic tag with empty div or fragment

### DIFF
--- a/src/ast-to-react.js
+++ b/src/ast-to-react.js
@@ -211,6 +211,11 @@ function assignDefined(target, attrs) {
 
 function mergeNodeChildren(node, parsedChildren) {
   const el = node.element
+  if (el === undefined) {
+    /* istanbul ignore next - `div` fallback for old React. */
+    const Fragment = React.Fragment || 'div'
+    return React.createElement(Fragment, null, null)
+  }
   if (Array.isArray(el)) {
     /* istanbul ignore next - `div` fallback for old React. */
     const Fragment = React.Fragment || 'div'
--- a/src/ast-to-react.js
+++ b/src/ast-to-react.js
@@ -176,7 +176,7 @@ function getNodeProps(node, key, opts, renderer, parent, index) {
       props.allowDangerousHtml = opts.allowDangerousHtml
       props.escapeHtml = opts.escapeHtml
       props.skipHtml = opts.skipHtml
-      props.element = mergeNodeChildren(node, parsedChildren)
+      props.element = node.element ? mergeNodeChildren(node, parsedChildren) : null
       break
     }
     default:
@@ -211,11 +211,6 @@ function assignDefined(target, attrs) {
 
 function mergeNodeChildren(node, parsedChildren) {
   const el = node.element
-  if (el === undefined) {
-    /* istanbul ignore next - `div` fallback for old React. */
-    const Fragment = React.Fragment || 'div'
-    return React.createElement(Fragment, null, null)
-  }
   if (Array.isArray(el)) {
     /* istanbul ignore next - `div` fallback for old React. */
     const Fragment = React.Fragment || 'div'
--- a/test/react-markdown.test.js
+++ b/test/react-markdown.test.js
@@ -894,9 +894,10 @@ test('should support formatting at the start of a GFM tasklist (GH-494)', () => 
   expect(actual).toEqual(expected)
 })
 
-test('should not crush on ending tag at start', () => {
+test('should not crash on weird `html-to-react` results', () => {
   const input = '<ruby><ruby></ruby></ruby>'
-  expect(() =>
-    renderHTML(<Markdown source={input} escapeHtml={false} astPlugins={[htmlParser()]} />)
-  ).not.toThrow()
+  const actual = renderHTML(<MarkdownWithHtml allowDangerousHtml children={input} />)
+  // Note: this is not conforming to how browsers deal with it.
+  const expected = '<p><ruby></ruby><ruby></ruby></p>'
+  expect(actual).toEqual(expected)
 })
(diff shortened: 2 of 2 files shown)


wooorm:
Hi there! It seems the coverage is not hitting your code, so it seems your test doesn’t do anything?

cerkiewny:
Let me have a look might be the case that my test is too simplistic.

cerkiewny:
Oh I haven't configured html parser which is causing the error, let me correct that.

cerkiewny:
I tracked it down some more, I think in my case it boils down to illegal html tag nesting, which, all in all isn't valid input, but I wouldn't expect my app to just to choke on that.

cerkiewny:
@wooorm any update on this one?

test/react-markdown.test.js
```diff
@@ -893,3 +893,10 @@ test('should support formatting at the start of a GFM tasklist (GH-494)', () =>
   const expected = '<ul><li><input type="checkbox" readonly=""/><em>a</em></li></ul>'
   expect(actual).toEqual(expected)
 })
+
+test('should not crush on ending tag at start', () => {
+  const input = '<ruby><ruby></ruby></ruby>'
+  expect(() =>
+    renderHTML(<Markdown source={input} escapeHtml={false} astPlugins={[htmlParser()]} />)
+  ).not.toThrow()
```
ChristianMurphy, line 901: @cerkiewny what is the actual resulting value?
Could that be included in the test?

ChristianMurphy (changes requested):

test/react-markdown.test.js
cerkiewny: to be honest I expect it just not to throw, I can add expectation that it is returning whatever it is returning now (which is <p><ruby></ruby><ruby></ruby></p>) but for me it isn't really obvious this is expected, let me know if I should add it to the test or leave it as is. To phrase my concerns otherwise I think even if it would change to something like <p><ruby></ruby></p> I don't think this test should stop passing.

I am happy to add it to the test though if you think it is appropriate change.

ChristianMurphy: I ask because `remark` and it's related project aim to match commonmark.
https://spec.commonmark.org/dingus/?text=%3Cruby%3E%3Cruby%3E%3C%2Fruby%3E%3C%2Fruby%3E%0A
which for this example would produce `<p><ruby><ruby></ruby></ruby></p>`

cerkiewny: I understand, the thing is, for now it produces the exception. And my fix was meant to just mitigate just that. I am not sure how to proceed, I can dive deeper into the parser and try and understand and fix the expectation mismatch however it will probably take a while. Meanwhile the output doesn't match expectation and the markdown parser simply throws. So we can either wait for the fix or merge as-is with follow up to fix the expectation. What do you think?

wooorm:
I’m assuming the root cause of this issue, as with many of the other open ones, is [`aknuds1/html-to-react`](https://github.com/aknuds1/html-to-react), which in it uses `htmlparser2`.

According to the HTML spec, [a ruby element can contain another ruby element, if that child is empty](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-ruby-element).
this can be confirmed in browsers by running: `document.body.innerHTML = '<ruby><ruby></ruby></ruby>'`, and inspecting the resulting dom: an empty ruby in a ruby.

It is a somewhat weird case, but for now we are crashing on it, and it would be better to not crash. And, meanwhile, as Christian says, it would be good to check the actual output and add a comment saying that it is currently incorrect output.

Finally, there is no reason to render a fragment/div. I’ll add a suggestion for where and how to solve this cleaner

test/react-markdown.test.js
```diff
@@ -893,3 +893,10 @@ test('should support formatting at the start of a GFM tasklist (GH-494)', () =>
...
+})
```
wooorm, line 902: (continuing from earlier conversation: https://github.com/remarkjs/react-markdown/pull/507/files#r525184845)

```suggestion
test('should not crash on weird `html-to-react` results', () => {
  const input = '<ruby><ruby></ruby></ruby>'
  const actual = renderHTML(<MarkdownWithHtml allowDangerousHtml children={input} />)
  // Note: this is not conforming to how browsers deal with it.
  const expected = '<p><ruby></ruby><ruby></ruby></p>'
  expect(actual).toEqual(expected)
})
```

Here is my suggestion: this uses the new properties (source -> children, escapeHtml -> allowDangerousHtml), and uses our component loaded with the html parser plugin. It adds a comment explaining why this test is needed.

src/ast-to-react.js
```diff
@@ -211,6 +211,11 @@ function assignDefined(target, attrs) {
 
 function mergeNodeChildren(node, parsedChildren) {
   const el = node.element
+  if (el === undefined) {
+    /* istanbul ignore next - `div` fallback for old React. */
+    const Fragment = React.Fragment || 'div'
+    return React.createElement(Fragment, null, null)
+  }
```
wooorm, line 218: I suggest removing this case: it’s not needed to create a fragment or div here

```suggestion
```

wooorm: Instead, on L179, how about this one?

```diff
-      props.element = mergeNodeChildren(node, parsedChildren)
+      props.element = node.element ? mergeNodeChildren(node, parsedChildren) : null
```

wooorm (changes requested):

cerkiewny:
@wooorm added changes as the fixup, seems github doesn't like force pushes with amends :)

wooorm (approved):

wooorm:
Thanks @cerkiewny!

@ChristianMurphy, could you review again?

ChristianMurphy (approved):
Thanks @cerkiewny !

wooorm:
thanks @cerkiewny, released!

cerkiewny:
Thanks!

PR #510, opened 2020-11-19
Fix umd browser build

doveccl:


wooorm (approved):

wooorm:
thanks @doveccl, released!

PR #512, opened 2020-11-21
allow passing in a parsed tree as a prop

entropitor:
If the user has already parsed the tree for other purposes, it can be useful to just pass in the parsed tree. This way the `renderers` logic etc can still be used while moving the parsing out

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -44,6 +44,7 @@ declare namespace ReactMarkdown {
       | null
     readonly renderers?: {[nodeType: string]: ElementType}
     readonly astPlugins?: PluggableList
+    readonly tree?: unist.Node
     readonly plugins?: PluggableList
     readonly unwrapDisallowed?: boolean
   }
--- a/src/react-markdown.js
+++ b/src/react-markdown.js
@@ -26,12 +26,15 @@ const ReactMarkdown = function ReactMarkdown(props) {
 
   const renderers = xtend(defaultRenderers, props.renderers)
 
-  const processor = unified()
-    .use(parse)
-    .use(props.plugins || [])
+  let tree = props.tree
+  if (!tree) {
+    const processor = unified()
+      .use(parse)
+      .use(props.plugins || [])
 
-  // eslint-disable-next-line no-sync
-  let tree = processor.runSync(processor.parse(src))
+    // eslint-disable-next-line no-sync
+    tree = processor.runSync(processor.parse(src))
+  }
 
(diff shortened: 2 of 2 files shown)


ChristianMurphy:
An interesting idea @entropitor. :thinking: 
An alternative would be passing the tree to https://github.com/remarkjs/remark-rehype and https://github.com/rehypejs/rehype-react directly.
Could you expand on why you see this as being a good fit for `react-markdown` specifically?

---

If you do seeing it as being a good fit at this level of abstraction a few additional considerations:
1. would it make sense to add a signal, that only `tree` or `children` can be passed? If both are passed one would be ignored, and from the documentation, it isn't clear which (it appears from implementation `tree` takes precedence)
2. a `tree` is `mdast` something which `plugins` can still process, would it make sense to pass the tree through plugins before rendering?
3. would it make sense to add a test case to cover the new logic?

entropitor:
It might be a solution but I don't see how those tools allow you to have custom renderers, which is something that react-markdown does really well.

1. Yes, I guess so, how would you signal this? 
2. I'm very new to the whole ecosystem, so I didn't know that was possible, I guess that makes sensee. I'm not sure how it should be done though
3. I guess that could make sense. But I guess the stuff above needs to be resolved first

ChristianMurphy:
> I don't see how those tools allow you to have custom renderers

https://github.com/syntax-tree/mdast-util-to-hast#optionshandlers and/or https://github.com/rehypejs/rehype-react#optionscomponents

> Yes, I guess so, how would you signal this?

One way would be through typings, using the `MutuallyExclusive` utility, similar to `source` and `children` are exclusive https://github.com/remarkjs/react-markdown/blob/22bb78747d768181cb9ea8711b5e13c3768921d8/index.d.ts#L82

> I'm very new to the whole ecosystem, so I didn't know that was possible, I guess that makes sensee. I'm not sure how it should be done though

to run just plugins/transformers, `unified` has ` run` method: https://github.com/unifiedjs/unified#description and https://github.com/unifiedjs/unified#processorrunnode-file-done

wooorm:
This to me feels like a rather complex (most folks don‘t know about syntax trees and how to make them) edge case (most folks wont need it). I think react-markdown instead should be simple (as simple as possible) and common.

And: the underlying tools can be used on their own, combined in all the ways you’d want. @entropitor Maybe this is an XY problem? You have a solution, but maybe we can help you better if we know about the problem?

entropitor:
I think it might be possible with those solutions then. I'll have to try it out soon! :D

entropitor:
I'll close this for now

Issue #529, opened 2021-01-14
Can't render a table


blb78:
### Can't render a table with gfm plugin
Hi ! 
I see that your demo's page can't render it too [see the table section](https://remarkjs.github.io/react-markdown/)

Maybe if you know why your demo is buggy it can help me. Because I can't reproduce on a fresh install.
And I don't why I can't render table...
I post this report on this repo because I see in your demo's page the bug. 
My bad if you think this is not the right place.

### Your environment
*   **OS**: Win10 pro
*   **Packages**:
    -  react-dom : 17.0.1
    -  react : 17.0.1
    -  react-markdown : 5.0.3
    -  remark-gfm : 1.0.0
*   **Env**:
    - nodejs : 14.15.1
    - npm : 6.14.8
    - firefox : 84.0.2

### Steps to reproduce
I can't reproduce !

On a fresh install it works

- npx create-react-app my-app --template typescript
- yarn add react-markdown
- yarn add remark-gfm

### Expected behavior
See a html table

### Actual behavior

See a inline markdown table, like your demo

ChristianMurphy:
duplicate of https://github.com/remarkjs/react-markdown/issues/526, https://github.com/remarkjs/react-markdown/issues/525, https://github.com/remarkjs/react-markdown/issues/524, https://github.com/remarkjs/react-markdown/issues/516, and https://github.com/remarkjs/react-markdown/issues/493

how to use tables (and the rest of GitHub Flavored Markdown) is documented in the readme https://github.com/remarkjs/react-markdown#use
As for the demo site, it does not include `remark-gfm` and therefore doesn't render tables.

Issue #526, opened 2020-12-22
Tables not working on website


lorensr:
### Subject of the issue

![image](https://user-images.githubusercontent.com/251288/102928091-226f3b00-4466-11eb-9582-5e2b403253dc.png)

### Your environment

*   **OS**: macOS 11.1
*   **Browser**: 

![image](https://user-images.githubusercontent.com/251288/102928021-08cdf380-4466-11eb-9a25-273f9bd7a2d7.png)

### Steps to reproduce

Visit https://remarkjs.github.io/react-markdown/

ChristianMurphy:
Tables, and all GFM have moved to a plugin.
You can see an example of usage here https://github.com/remarkjs/react-markdown#use
and the note in the release notes here: https://github.com/remarkjs/react-markdown/blob/main/changelog.md#500---2020-10-19

The website accurately reflects that by default, react-markdown follows the CommonMark standard, which does not have Tables.

lorensr:
I think it would be helpful to make note of that on the website

wooorm:
@lorensr I think you’re right that the demo on the website should use GFM, and describe that tables/strikethrough etc are enabled through a plugin. A new issue or even a PR would be awesome!

chloerice:
> Tables, and all other GFM have moved to a plugin. You can see an example of usage with tables here: https://github.com/remarkjs/react-markdown#use and the note in the release notes here: https://github.com/remarkjs/react-markdown/blob/main/changelog.md#500---2020-10-19
> 
> The website accurately reflects that by default, react-markdown follows the CommonMark standard, which does not have Tables.

[Here's the direct link](https://github.com/remarkjs/react-markdown#use-a-plugin) to the example referencing the `remark-gfm` plugin illustrating table parsing

georgeportillo:
+1

Issue #525, opened 2020-12-14
Table is not shown properly


ppascualv:
### Subject of the issue

As you can see in the example attached to the readme https://remarkjs.github.io/react-markdown/ tables are not shown properly this is something that has been broken recently.

### Expected behavior

It should be shown as a table.

### Actual behavior

It is shown as inline text

ChristianMurphy:
Table support is explained in the changelog and in the readme.
As well as explained in the previous questions asked
https://github.com/remarkjs/react-markdown/issues/524
https://github.com/remarkjs/react-markdown/issues/516
https://github.com/remarkjs/react-markdown/issues/493

Issue #524, opened 2020-12-13
Checkboxes, Tables & Strikethrough don't work anymore 


dbuchet:
### Subject of the issue
Updating to 5.0.3 breaks several features:
- Checkboxes are not rendered
- Tables are not rendered
- Strikethrought texts are not rendered

### Your environment

*   **OS**: All
*   **Packages**: react-markdown 5.0.3, React17, CRA 4
*   **Env**: all browsers

### Steps to reproduce
https://codesandbox.io/s/boring-chatterjee-6vt5s
Happens also on playground here: https://remarkjs.github.io/react-markdown/

Thanks!

ChristianMurphy:
These features are part of GitHub Flavored Markdown (GFM), which is now a plugin.
This is documented in the changelog https://github.com/remarkjs/react-markdown/blob/main/changelog.md#upgrade-remark-parse
and an example of enabling these features is shown in the readme https://github.com/remarkjs/react-markdown#use

wooorm:
See the readme, the change log, or recently closed issues

dbuchet:
Oh sorry didn't notice. Thanks for your quick reply

PR #530, opened 2021-01-14
docs: remove github flavored table from example/demo of commonmark

ChristianMurphy:
including an example of GitHub flavored markdown in the CommonMark demo leads to confusion (see https://github.com/remarkjs/react-markdown/issues/529, https://github.com/remarkjs/react-markdown/issues/524, https://github.com/remarkjs/react-markdown/issues/525, https://github.com/remarkjs/react-markdown/issues/526).
This removes GFM from the commonmark demo to reduce confusion.

An alternative could be to add a toggle to enable/disable `remark-gfm` on the demo site.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -40,11 +40,18 @@ Pretty neat, eh?
 
 ## Tables?
 
+Use [\`remark-gfm\`](https://github.com/remarkjs/react-markdown#use) to support tables, strikethrough, tasklists, and literal URLs.
+Theses features **do not work by default**.
+
 | Feature   | Support |
 | :-------: | ------- |
-| tables    | ✔ |
-| alignment | ✔ |
-| wewt      | ✔ |
+| tables    | \`remark-gfm\` |
+
+~~strikethrough~~
+
+- [ ] task list
+
+https://example.com
 
 ## More info?
 


wooorm:
Definitely something to fix, thinking about this some more, I think it’s good to instead *document* that it’s missing:

```markdown
Tables **do not work by default**: use [`remark-gfm`](url) to support tables, strikethrough, tasklists, and literal URLs:

| the | table |
```

And perhaps have a checkbox next to `HTML Mode` with ``Use [`remark-gfm`](url)``?

ChristianMurphy:
I agree, including some details on GFM support in the demo and having a toggle could be beneficial.
It is a bit more involved of an update as well, maybe we hold of this and do the demo/docs update as part of v6 rehype update?

wooorm:
The toggle is perhaps a bit much for now, but could we land this already with something along the lines of my [suggested change](https://github.com/remarkjs/react-markdown/pull/530#issuecomment-760323408) (documenting it instead of removing it)?

ChristianMurphy:
Sure, that could be an option, what do you think of 3a6095a ?

wooorm (approved):

wooorm:
Marvelous!

PR #531, opened 2021-01-14
Fix typo

sergeymirasov:


ChristianMurphy (changes requested):
the current comment is correct
for example try
````markdown
```
a
```
````
in https://remarkjs.github.io/react-markdown
it is wrapped in `<pre>` and `<code>`

sergeymirasov:
Excuse me. And thanks for explaining

PR #543, opened 2021-02-06
Fix `uri-transformer.js` link on README

kenzoukenzou:
close #542 
Fix `uri-transformer.js` link on README.

PR #544, opened 2021-02-08
Fix broken link in README

chartgerink:
👋 I was working through the docs and noticed this link was broken. I think this may fix it, if I understand correctly.

Cheers,
Chris


wooorm:
Thank you!

Interestingly someone else also ran into it a couple days ago, and there was already a PR open for it, so I merged that: https://github.com/remarkjs/react-markdown/pull/543

PR #547, opened 2021-02-15
Refactor more internals to be remark plugins

wooorm:
This is a first step in making a switch to rehype in internals possible.

* Remove `astPlugins`
* Add temporary `htmlParser` prop to pass the htmlparser2 based HTML parser
  (plan: to use `rehype-raw`)
* Fix a couple visitors


src/remark-filter-nodes.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    if (remove) {
+      let parameters = [index, 1]
+
+      if (options.unwrapDisallowed && node.children) {
+        parameters = parameters.concat(node.children)
+      }
+
+      splice.apply(parent.children, parameters)
```
ChristianMurphy, line 42: would this be the same as

```suggestion
      parent.children = [...parent.children, ...parameters]
```

src/plugins/html-parser.js
```diff
@@ -178,15 +97,78 @@ function parsedHtml(fromNode, toNode, parent) {
   }
 }
 
-module.exports = function getHtmlParserPlugin(config, props) {
-  if (props && typeof config.children !== 'undefined') {
+module.exports = function getHtmlParserPlugin(options, _2) {
```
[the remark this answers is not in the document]
wooorm: Doesn’t really matter: it’s not an accepted PR, but it’s used to sniff if the function is called wrongly.
Anyway, want to get rid of this and replace it with rehype-raw

src/remark-filter-nodes.js
wooorm: Yeah, although I’m assuming splice is faster though because it mutates (see micromark things a while ago).
Anyway, this code was already here, I copied it in from a now-removed file. I’m not sure what the baseline for react-markdown and es version is, because there’s also `xtend` instead of `Object.assign`.
Maybe something for the ESM move?

ChristianMurphy: > Maybe something for the ESM move?

fine by me

ChristianMurphy (approved):

PR #548, opened 2021-02-21
Remove old polyfills

wooorm:
* Add restriction to ES 2015
* Use `Object.assign` instead of `xtend`
* Require `React.Fragment`
* Remove hacks needed for React 15

### DIFF
--- a/src/ast-to-react.js
+++ b/src/ast-to-react.js
@@ -1,7 +1,6 @@
 'use strict'
 
 const React = require('react')
-const xtend = require('xtend')
 const ReactIs = require('react-is')
 
 function astToReact(node, options, parent = {}, index = 0) {
@@ -128,7 +127,7 @@ function getNodeProps(node, key, opts, renderer, parent, index) {
     case 'linkReference':
       assignDefined(
         props,
-        xtend(ref, {
+        Object.assign({}, ref, {
           href: opts.transformLinkUri ? opts.transformLinkUri(ref.href) : ref.href
         })
       )
@@ -180,7 +179,7 @@ function getNodeProps(node, key, opts, renderer, parent, index) {
     default:
       assignDefined(
         props,
-        xtend(node, {
+        Object.assign({}, node, {
           type: undefined,
           position: undefined,
           children: undefined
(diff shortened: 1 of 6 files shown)


src/ast-to-react.js
```diff
@@ -128,7 +127,7 @@ function getNodeProps(node, key, opts, renderer, parent, index) {
     case 'linkReference':
       assignDefined(
         props,
-        xtend(ref, {
+        Object.assign({}, ref, {
           href: opts.transformLinkUri ? opts.transformLinkUri(ref.href) : ref.href
         })
```
ChristianMurphy, line 132: with the broader syntax allowance, would it make sense to use:

```suggestion
        {
          ...ref,
          href: opts.transformLinkUri ? opts.transformLinkUri(ref.href) : ref.href
        }
```
?

Or even

```suggestion
        {
          ...ref,
          href: opts.transformLinkUri?.(ref.href) ?? ref.href
        }
```

ChristianMurphy (approved):

src/ast-to-react.js
wooorm: The nullish stuff is way too new I think. That’s going to trip up people.

For the spread: I’m guessing spread has the same support as `Object.assign`? 🤔 If so, that’s fine, if not, I’d rather wait.

I added `Object.assign` because `.includes` is also used in the code in other places. So this is more me streamlining the current “version” than me updating things. More changes can come with the ESM move?

ChristianMurphy: > For the spread: I’m guessing spread has the same support as Object.assign?

Not quite the same, below is the full support matrix

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#browser_compatibility

Picking Node for example `object.assign` was added in Node 4, object spread in Node 8.

wooorm: Seems that especially spread in objects is pretty late. Hmm, I’d prefer `Object.assign` for now, and modernizing this project in a next major over the summer or so

PR #571, opened 2021-04-14
Remove compiled library

wooorm:
I don’t think we need a prebuilt `lib/`; the source code is usable as-is

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,2 @@
 coverage/
-lib/
 react-markdown.min.js
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 coverage/
-lib/
 node_modules/
 *.d.ts
 *.log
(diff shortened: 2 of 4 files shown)


ChristianMurphy (approved):
The main difference appears to be that we'd lose TS as a transpile to guarantee that output is ES2015.
Though scanning through the files it appears they are already ES2015.

wooorm:
How is TS being used to transpile?

ChristianMurphy:
Perhaps I misread the tsconfig, but my impression was that it read the source dir and copied/transpiled to the lib?

wooorm:
I don’t think so, `build:lib` was doing it:

```
"build:lib": "rimraf lib/ && babel src/ --quiet --copy-files -d lib/",
```

I think TS was checking *both* (might be a mistake I made recently), but as they were basically copies, that didn’t do much

PR #575, opened 2021-04-18
Remove trailing \n in pre.code

shfshanyue:
every `pre.code` tag have one redundant blank line

**Expected**

```
line 1 of code
line 2 of code
line 3 of code
```

**Actual**

```
line 1 of code
line 2 of code
line 3 of code

```

### DIFF
--- a/src/ast-to-react.js
+++ b/src/ast-to-react.js
@@ -171,7 +171,11 @@ function childrenToReact(context, node) {
     if (child.type === 'element') {
       children.push(toReact(context, child, childIndex, node))
     } else if (child.type === 'text') {
-      children.push(child.value)
+      if (node.tagName === 'code' && !node.properties.inline) {
+        children.push(child.value.replace(/\n$/, ''))
+      } else {
+        children.push(child.value)
+      }
     }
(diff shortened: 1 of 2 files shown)


wooorm:
That extra line feed is supposed to be there: https://spec.commonmark.org/dingus/?text=%60%60%60%0Aasd%0A%60%60%60 (see HTML tab).

I think `react-syntax-highlighter` is adding *another* one? Can you confirm you’re using that, and seeing a problem with it? Or is it something else?

wooorm:
Pushed a change to the example which removes that EOF EOL before passing it to `react-syntax-highlighter`

shfshanyue:
I test a case for codeblock，it's ok, `react-syntax-highlighter` is adding another one. thanks

``` js
test('should render codeblocks', () => {
  const input = '```\nline 1 of code\nline 2 of code\nline 3 of code\n```'
  const actual = renderHTML(<Markdown children={input} />)
  const expected =
    '<pre><code>line 1 of code\nline 2 of code\nline 3 of code\n</code></pre>'
  expect(actual).toEqual(expected)
})

```

Issue #576, opened 2021-04-21
React Markdown 6.0.0 with Remark GFM 1.0.0 - Warning: validateDOMNesting(...): Whitespace text nodes cannot appear as a child of <table>. Make sure you don't have any extra whitespace between tags on each line of your source code.


fdcds:
### Subject of the issue

When rendering:
```
  <ReactMarkdown
    plugins={[gfm]}
  >
    {`| table | md |
| --- | --- |
| row1 | more |
| row2 | text |`}
  </ReactMarkdown>
```

with React v17.0.2, React Markdown v6.0.0, Remark Parse 9.0.0, Remark Rehype 8.1.0, Remark GFM 1.0.0, Firefox 87.0 will display warnings in the web developer console.

### Your environment

*   **OS**: Fedora 33, KDE Spin
*   **Packages**: React 17.0.2, React Markdown 6.0.0, Remark Parse 9.0.0, Remark Rehype 8.1.0, Remark GFM 1.0.0
*   **Env**: NodeJS 15.11.0, NPM 7.6.0, Firefox 87.0

### Steps to reproduce

Minimal reproduction: https://codesandbox.io/s/react-markdown-debug-forked-otbbg

### Expected behavior

Table should render without React warnings.

### Actual behavior

```
Warning: validateDOMNesting(...): Whitespace text nodes cannot appear as a child of <table>. Make sure you don't have any extra whitespace between tags on each line of your source code.
table
div
ReactMarkdown@http://localhost:3000/app/static/js/vendors~main.chunk.js:154250:42
[...]

Warning: validateDOMNesting(...): Whitespace text nodes cannot appear as a child of <thead>. Make sure you don't have any extra whitespace between tags on each line of your source code.
thead
table
div
ReactMarkdown@http://localhost:3000/app/static/js/vendors~main.chunk.js:154250:42
[...]

Warning: validateDOMNesting(...): Whitespace text nodes cannot appear as a child of <tr>. Make sure you don't have any extra whitespace between tags on each line of your source code.
tr
thead
table
div
ReactMarkdown@http://localhost:3000/app/static/js/vendors~main.chunk.js:154250:42
[...]

Warning: validateDOMNesting(...): Whitespace text nodes cannot appear as a child of <tbody>. Make sure you don't have any extra whitespace between tags on each line of your source code.
tbody
table
div
ReactMarkdown@http://localhost:3000/app/static/js/vendors~main.chunk.js:154250:42
[...]
```

[...]

PR #577, opened 2021-04-21
Fix whitespace in table elements

fdcds:
React Markdown 6.0.0 with Remark Parse 9.0.0, Remark Rehype 8.1.0, Remark
GFM 1.0.0 and React 17.0.2 would cause warnings in the developer console:

> Warning: validateDOMNesting(...): Whitespace text nodes cannot appear as a
> child of `<table>`. Make sure you don't have any extra whitespace between
> tags on each line of your source code.

In fact `<table>`, `<tbody>`, `<thead>`, `<tr>` all do not permit any text
elements to appear directly inside them.  See e.g.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table

Hence we strip any text element from the children of these elements
unconditionally.

Closes: https://github.com/remarkjs/react-markdown/issues/576

### DIFF
--- a/src/ast-to-react.js
+++ b/src/ast-to-react.js
@@ -172,11 +172,13 @@ function childrenToReact(context, node) {
       children.push(toReact(context, child, childIndex, node))
     } else if (child.type === 'text') {
       // Text elements are not permitted as children of table:
-      if (node.tagName !== 'table' &&
-          node.tagName !== 'thead' &&
-          node.tagName !== 'tbody' &&
-          node.tagName !== 'tfoot' &&
-          node.tagName !== 'tr') {
+      if (
+        node.tagName !== 'table' &&
+        node.tagName !== 'thead' &&
+        node.tagName !== 'tbody' &&
+        node.tagName !== 'tfoot' &&
+        node.tagName !== 'tr'
+      ) {
         children.push(child.value)
       }
     }
--- a/src/ast-to-react.js
+++ b/src/ast-to-react.js
@@ -171,7 +171,14 @@ function childrenToReact(context, node) {
     if (child.type === 'element') {
       children.push(toReact(context, child, childIndex, node))
     } else if (child.type === 'text') {
-      children.push(child.value)
+      // Text elements are not permitted as children of table:
+      if (node.tagName !== 'table' &&
+          node.tagName !== 'thead' &&
+          node.tagName !== 'tbody' &&
+          node.tagName !== 'tfoot' &&
+          node.tagName !== 'tr') {
+        children.push(child.value)
+      }
     }
     // @ts-ignore `raw` nodes are non-standard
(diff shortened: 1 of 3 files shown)


src/ast-to-react.js
```diff
@@ -171,7 +171,13 @@ function childrenToReact(context, node) {
     if (child.type === 'element') {
       children.push(toReact(context, child, childIndex, node))
     } else if (child.type === 'text') {
-      children.push(child.value)
+      if (node.tagName !== 'table' &&
```
fdcds, line 174:
```suggestion
      // Text elements are not permitted as children of table:
      if (node.tagName !== 'table' &&
```

wooorm (commented):
Hi again, thanks for tackling this!

Could you refactor this a bit?
I think the linters will probably format an alternative too.
You can run `npm test` locally to format, check, and test everything.

I suggest the following changes:

*   Can you create a list at the top of the file: `var tableElements = [...]`, and here use it as such `tableElements.includes(node.tagName)`
*   Make sure that *only* `node.value === '\n'` is filtered out (or, use `/^[ \t\n\f\r]+$/` to get fancy)
*   Add a comment to this issue, so it’s clear why this “hack” is in place

fdcds:
> You can run `npm test` locally to format, check, and test everything.

Thanks for that hint!

>     * Can you create a list at the top of the file: `var tableElements = [...]`, and here use it as such `tableElements.includes(node.tagName)`

This got transformed to `Set([...]).has(...)` by npm test (or by IntelliJ?). It made sense, so I kept it.

>     * Make sure that _only_ `node.value === '\n'` is filtered out (or, use `/^[ \t\n\f\r]+$/` to get fancy)

I originally filtered just that, but for testing purposes replaced the white space text nodes with `"SOMETEXT"` and then React complained about text nodes being not permitted there, so I decided I'd remove them always.

Following your comment I now reverted to removing white space only. It makes sense to be more conservative, so I think it's fine.

In general I wonder: Why is there white space in the first place? Is this a bug in `remark` or `react-markdown` (or in case of tables: in `remark-gfm`)?

wooorm:
> Following your comment I now reverted to removing white space only. It makes sense to be more conservative, so I think it's fine.

Yeah, `SOMETEXT` is also incorrect, so good that React warns for that, but in this case I think it’s safer to *keep* warning when users somehow do inject text in between tables.

> In general I wonder: Why is there white space in the first place? Is this a bug in remark or react-markdown (or in case of tables: in remark-gfm)?

Good Q! It doesn’t really matter here, but generally CommonMark (and GFM) add `\n` between blocks. The tooling we use is also used in non-React places, so it also does that.
I don’t understand why React hasn’t allowed this though — it does allow `\n` in many other places. So it’s more of a React thing (this project) to remove them

wooorm (approved):

wooorm:
thanks, released!

PR #582, opened 2021-04-28
Add string type for `className`s

kikagaku:
Hello.
Thank you for the very nice package.
Thanks to react-markdown, my development is going very smoothly.

I added the `className` to the `CodeComponent` type because it is not defined in the script, while it is in `README.md`.
I have confirmed that it works on my environment, so I would be happy if you could adopt it.

Thank you for your consideration.

スクリーンショット 2021-04-28 13 38 00

### DIFF
--- a/src/ast-to-react.js
+++ b/src/ast-to-react.js
@@ -77,7 +77,8 @@ exports.hastChildrenToReact = childrenToReact
  *
  * @typedef {keyof IntrinsicElements} ReactMarkdownNames
  *
- * @typedef {Object.<string, unknown>} ReactBaseProps
+ * @typedef {Object} ReactBaseProps
+ * @property {string?} className
  *
  * To do: is `data-sourcepos` typeable?
  *
@@ -94,7 +95,7 @@ exports.hastChildrenToReact = childrenToReact
  * @returns {ReactNode}
  *
  * @callback CodeComponent
- * @param {ReactBaseProps & ReactMarkdownProps & {inline?: boolean, className?: string}} props
+ * @param {ReactBaseProps & ReactMarkdownProps & {inline?: boolean}} props
  * @returns {ReactNode}
  *
  * @callback HeadingComponent
--- a/src/ast-to-react.js
+++ b/src/ast-to-react.js
@@ -94,7 +94,7 @@ exports.hastChildrenToReact = childrenToReact
  * @returns {ReactNode}
  *
  * @callback CodeComponent
- * @param {ReactBaseProps & ReactMarkdownProps & {inline?: boolean}} props
+ * @param {ReactBaseProps & ReactMarkdownProps & {inline?: boolean, className?: string}} props
  * @returns {ReactNode}
  *
  * @callback HeadingComponent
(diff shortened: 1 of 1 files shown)


ChristianMurphy (approved):
Thanks @kikagaku! :bow: 

---

@wooorm general question, would it make sense to move `className` up a level to `ReactBaseProps`?

wooorm:
I’m not 100% this is the way to go? E.g., in JSX one can do `<MyComponent className={1} />`?

I’ve thought about typing hast more. Having elements and their expected attributes typed as detailed as possible. But it remains that people can use custom props. It seems like quite the endeavour though.

Murderlon:
> I’m not 100% this is the way to go? E.g., in JSX one can do `<MyComponent className={1} />`?

Technically you can, but when you start using TS the [React types also just use string](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v16/index.d.ts#L1750).

> I’ve thought about typing hast more. Having elements and their expected attributes typed as detailed as possible. But it remains that people can use custom props. It seems like quite the endeavour though.

Are you talking about typing hast elements outside of react-markdown? Because the ones for components here look already typed.

wooorm:
> Technically you can, but when you start using TS the React types also just use string.

Ah, okay. Let’s stick with that.

> Are you talking about typing hast elements outside of react-markdown? Because the ones for components here look already typed.

Indeed, `@types/hast`; adding info on properties that resides in https://github.com/wooorm/property-information. But that’s a separate issue.

---

@kikagaku I think I agree with @ChristianMurphy’s earlier comment: could you add `className?` to `ReactBaseProps`?

Murderlon (approved):

wooorm:
Hmm, did you test it? Not 100% sure but I think in JSDoc either the stuff in curlies, or following properties work 🤔

src/ast-to-react.js
```diff
@@ -77,7 +77,8 @@ exports.hastChildrenToReact = childrenToReact
  *
  * @typedef {keyof IntrinsicElements} ReactMarkdownNames
  *
- * @typedef {Object.<string, unknown>} ReactBaseProps
+ * @typedef {Object} ReactBaseProps
+ * @property {string?} className
```
ChristianMurphy, line 81:
```suggestion
 * @typedef {{ [key: string]: unknown, className?: string }} ReactBaseProps
```

https://www.typescriptlang.org/play?useJavaScript=true#code/PQKhCgAIUgBAXAngBwKYBNUDNIG9eQDaA1qogFyQDO8ATgJYB2A5gLqUCujxjA9gO6MANJADGAGwCGVKgDlJAW1QB+SjQYtIAXy2QAogA9Fycaighg4cKAjQ4SNHkPHTW85dG9GNSPFQ+AXjwtIA

wooorm: @kikagaku You can click “commit suggestion” on these suggestions to add then. I just did that for you!

wooorm (approved):

wooorm:
Released!

PR #587, opened 2021-05-04
fix react-syntax-highlighter example in README

rohitbakoliya:
In a given `react-syntax-highlighter` example, it skips inline code rendering, so I fixed it.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -252,7 +252,12 @@ import {render} from 'react-dom'
 const components = {
   code({node, inline, className, children, ...props}) {
     const match = /language-(\w+)/.exec(className || '')
-    return !inline && match ? (
+    if(inline) {
+      return (
+        <SyntaxHighlighter style={dark} useInlineStyles={true} PreTag="span" children={String(children).replace(/\n$/, '')} {...props} />
+      )
+    }
+    return match ? (
       <SyntaxHighlighter style={dark} language={match[1]} PreTag="div" children={String(children).replace(/\n$/, '')} {...props} />
     ) : (
       <code className={className} {...props} />


wooorm:
The goal of that example is to specifically only highlight code with a language flag

wooorm:
Closing this as it’s intended. But thanks for offering to help!

ambroiseRabier:
I agree with @rohitbakoliya , thanks you for fixing it. Most people reading the documentation will want to know how to add basic syntax highlighting. Without the bonus feature "delete content of inline code elements".

I am surprised there is so little clatter about that part of the documentation.

At minimum, the documentation should explicitly state something like "This example bellow show how to delete the content of inline code snippet". As for now it only state that it does highlighting:

> This example shows how you can overwrite the normal handling of an element by passing a component. In this case, we apply syntax highlighting with the seriously super amazing react-syntax-highlighter by @conorhastings:

wooorm:
What is being deleted?

ambroiseRabier:
`inline code like this one doesn't display`, I have confused something, because it seem I can actually fix it by doing:

```tsx
        <code className={className} {...props}>
          {children}
        </code>
```

instead of
```tsx
<code className={className} {...props}/>
```

wooorm:
Before, `children` was not removed from `props`, so you are right that this is now a bug that I was unaware off, and that’s something that should be fixed. Thanks for raising, however, this PR you are commenting on does different things, too. Would you be interested in PRing your fix?

The `PreTag` and `children.replace` stuff is because of `react-syntax-highlighter`.

ambroiseRabier:
I can make PR, I removed `children={String(children).replace(/\n$/, '')}` and it doesn't seem to change anything, that is why I am asking.

wooorm:
See https://github.com/remarkjs/react-markdown/pull/575.

Could you point your PR to our project? It looks like you pointed it to your own fork

ambroiseRabier:
Ah sorry, thanks for pointing it out. The PR should be pointing to this project now.

wooorm:
Thanks, fixed!

PR #588, opened 2021-05-05
Include sourcePosition with rehype-raw

aslakhellesoy:
This PR fixes a bug where `sourcePosition` would not be set when `rehype-raw` is used as a rehype plugin.


wooorm:
Nice, thanks!

I landed it slightly differently, so that actual vfiles are passed to all plugins (which often expect them, not just a `{contents}` object).

wooorm:
Released!

aslakhellesoy:
Thanks for the quick turnaround @wooorm and thanks for a great lib!

PR #591, opened 2021-05-07
add `ImageComponent` into types

yoshizaki91:
Hello.
Thanks again for developing such a great package.

When using `components`, there was no property set for the `img` tag, so I created an `ImageComponent` type and applied it to the `img` tag.
I'm not sure if I should add `width` and `height` as well, so please add those as well if needed.

The test has of course passed, plus it has been verified to work in my environment.

Thank you for your consideration.

### DIFF
--- a/src/ast-to-react.js
+++ b/src/ast-to-react.js
@@ -77,7 +77,8 @@ exports.hastChildrenToReact = childrenToReact
  *
  * @typedef {keyof IntrinsicElements} ReactMarkdownNames
  *
- * @typedef {Object.<string, unknown>} ReactBaseProps
+ * @typedef {Object} ReactBaseProps
+ * @property {string?} className
  *
  * To do: is `data-sourcepos` typeable?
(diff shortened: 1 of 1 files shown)


Murderlon (approved):
Hi @yoshizaki91, this looks good to me and I think width and height types are not needed here.

Minor: perhaps these types resolve [this `ts-ignore`](https://github.com/remarkjs/react-markdown/blob/fb5da96e8033ac870486a9c6dfba49dbcb8c2a5d/src/ast-to-react.js#L308)?

yoshizaki91:
thanks comments.
I agree with your opinion and close this issue.

Thank you for your consideration.

Murderlon:
@yoshizaki91 maybe a miscommunication but we were happy to accept your change, there was merely a comment to see if some `ts-ignore` could be removed. Would you be able to reopen and check if that `ts-ignore` can be removed?

PR #603, opened 2021-06-12
Add `ImageComponent` into types.

SalahAdDin:
Almost the same like this: https://github.com/remarkjs/react-markdown/pull/591
Attend to fix this: https://github.com/remarkjs/remark/discussions/744#discussioncomment-863903

### DIFF
--- a/src/ast-to-react.js
+++ b/src/ast-to-react.js
@@ -101,6 +101,10 @@ exports.hastChildrenToReact = childrenToReact
  * @param {ReactBaseProps & ReactMarkdownProps & {level: number}} props
  * @returns {ReactNode}
  *
+ * @callback ImageComponent
+ * @param {ReactBaseProps & ReactMarkdownProps & {src: string, alt: string, title?: string}} props
+ * @returns {ReactNode}
+ *
  * @callback LiComponent
  * @param {ReactBaseProps & ReactMarkdownProps & {checked: boolean|null, index: number, ordered: boolean}} props
  * @returns {ReactNode}
@@ -121,6 +125,10 @@ exports.hastChildrenToReact = childrenToReact
  * @param {ReactBaseProps & ReactMarkdownProps & {depth: number, ordered: false}} props
  * @returns {ReactNode}
  *
+ * @callback VideoComponent
+ * @param {ReactBaseProps & ReactMarkdownProps & {url: string}} props
+ * @returns {ReactNode}
+ *
  * @typedef {Object} SpecialComponents
  * @property {CodeComponent|ReactMarkdownNames} code
  * @property {HeadingComponent|ReactMarkdownNames} h1
@@ -129,12 +137,14 @@ exports.hastChildrenToReact = childrenToReact
  * @property {HeadingComponent|ReactMarkdownNames} h4
  * @property {HeadingComponent|ReactMarkdownNames} h5
  * @property {HeadingComponent|ReactMarkdownNames} h6
+ * @property {ImageComponent|ReactMarkdownNames} img
  * @property {LiComponent|ReactMarkdownNames} li
  * @property {OrderedListComponent|ReactMarkdownNames} ol
  * @property {TableCellComponent|ReactMarkdownNames} td
  * @property {TableCellComponent|ReactMarkdownNames} th
  * @property {TableRowComponent|ReactMarkdownNames} tr
  * @property {UnorderedListComponent|ReactMarkdownNames} ul
+ * @property {VideoComponent|ReactMarkdownNames} video
  *
  * @typedef {Record<Exclude<ReactMarkdownNames, keyof SpecialComponents>, NormalComponent|ReactMarkdownNames>} NormalComponents
  * @typedef {Partial<NormalComponents & SpecialComponents>} Components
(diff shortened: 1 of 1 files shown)


src/ast-to-react.js
```diff
@@ -121,6 +125,10 @@ exports.hastChildrenToReact = childrenToReact
  * @param {ReactBaseProps & ReactMarkdownProps & {depth: number, ordered: false}} props
  * @returns {ReactNode}
  *
+ * @callback VideoComponent
+ * @param {ReactBaseProps & ReactMarkdownProps & {url: string}} props
```
ChristianMurphy, line 129: `url` isn't a `<video>` attribute :thinking: 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#attributes

It may make sense to add attributes supported by the tag, like `src`, `autoplay`, `muted`, `poster`, and `loop`.

ChristianMurphy (changes requested):
Thanks @SalahAdDin!

src/ast-to-react.js
SalahAdDin: Done!

wooorm:
@ChristianMurphy Are we going to type all possible elements (i.e., `video` does not often come from markdown).

---

If we’re adding `video`, then `autoplay` should be `autoPlay` I believe in React.

---

Alternatively, can we allow `NormalComponent` to have unknown keys in `props`?

ChristianMurphy:
> Are we going to type all possible elements (i.e., video does not often come from markdown).

If we can reflect the attributes from the React typings, perhaps using `IntrinsicElements`.
If we have to manually type the elements, probably not.

Murderlon (commented):
I think the `ImageComponent` is a good addition because it comes from markdown but I'm not sure if we should only add the `VideoComponent` from all the other HTML elements in this PR. Perhaps explore if we want that in a different PR?

SalahAdDin:
> > Are we going to type all possible elements (i.e., video does not often come from markdown).
> 
> If we can reflect the attributes from the React typings, perhaps using `IntrinsicElements`.
> If we have to manually type the elements, probably not.

What's could we do then for the `VideoComponent`?

ChristianMurphy:
@SalahAdDin maybe?
It would take some experimentation, but it may be possible to reflect the types from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/251ef8f61b7bd0c292d596177e68fd154c96e9c5/types/react/index.d.ts#L3060-L3240 to provide typings for all intrinsic elements in HTML and SVG.

wooorm:
@SalahAdDin Is fiddling with React types and updating what we have here to include all that something you could try and do?
To me, it sounds quite hard, and then I’d rather leave typing all of HTML—including videos—to a different PR (which perhaps @ChristianMurphy has ideas on?), and merge `img[alt]` for now

SalahAdDin:
> @SalahAdDin Is fiddling with React types and updating what we have here to include all that something you could try and do?
> To me, it sounds quite hard, and then I’d rather leave typing all of HTML—including videos—to a different PR (which perhaps @ChristianMurphy has ideas on?), and merge `img[alt]` for now

Well, the problem with embbebing `html` for videos is we have no so control to use other components to render them, and also.

We also found this way to render them since the **Strapi** _WYSIWYG_ editor is not so powerful to embed html and properly translate it to markdown.

SalahAdDin:
@ChristianMurphy can you lead me about the problem on tests?

ChristianMurphy:
@SalahAdDin try pulling in latest changes from `main` branch, it may be an issue which has already been resolved there.
`git pull https://github.com/remarkjs/react-markdown main`

SalahAdDin:
> @SalahAdDin try pulling in latest changes from `main` branch, it may be an issue which has already been resolved there.
> `git pull https://github.com/remarkjs/react-markdown main`

Done!

ChristianMurphy:
Thanks @SalahAdDin!
Thoughts on https://github.com/remarkjs/react-markdown/pull/603#pullrequestreview-682677660? :thought_balloon:

SalahAdDin:
> Thanks @SalahAdDin!
> Thoughts on [#603 (review)](https://github.com/remarkjs/react-markdown/pull/603#pullrequestreview-682677660)?

Done! I will open a new Pull request for video component, which is vital for our project.

ChristianMurphy (approved):
Thanks @SalahAdDin!

wooorm (approved):

wooorm:
@ChristianMurphy I guess this is also solved by GH-624?

ChristianMurphy:
Potentially.
Though `react-markdown` currently makes a distinction between `NormalComponent` (plain HTML) and `SpecialComponent` (HTML from markdown).
This being one of the latter.
With #624 is this distinction no longer needed?

wooorm:
`SpecialComponents` exists because they receive custom properties from `react-markdown`. `img` is not one of them

ChristianMurphy:
Gotcha, in that case, yes this is covered by #624

SalahAdDin:
> Gotcha, in that case, yes this is covered by #624

So, this is not required anymore, right?

wooorm:
Indeed. Sorry for that and thanks for your work.
I’ve removed other unneeded HTML info in [this commit](https://github.com/remarkjs/react-markdown/commit/b90ab75afb8742a9ff52a75b5bd4905fd0cade95) and added you as a contributor on it.

PR #604, opened 2021-06-14
Fix syntax highlighting example in readme.md

ambroiseRabier:
Fixed SyntaxHighlighter example in readme.md not rendering inline code, see See remarkjs#587


ChristianMurphy (approved):

wooorm (approved):

wooorm:
Thank you!

PR #606, opened 2021-06-15
Updated h1,h2,etc to headings

Mozart409:
I added a documentation update to reflect the new changes in version 6 regarding customRenders. This small update adds a note to the "h1,h2,etc" section linking to the general changelog to version 6
[Source](https://github.com/remarkjs/.github/blob/main/contributing.md)


wooorm:
uh, isn’t it the inverse? You used to use `heading`, and now it’s `h1` ... `h6`?

wooorm:
I believe this suggestion is incorrect, and were it fixed, I don’t think the readme is a place to duplicate the changelog.

PR #615, opened 2021-07-09
add peerDependenciesMeta in package.json

### LINKED ISSUE #614: @types/react peer dependency causes npm/yarn warnings
### Subject of the issue

We don't use TypeScript in our project and after updating to the latest react-markdown version we started getting the following warning during the dependencies installation:

`warning " > react-markdown@6.0.2" has unmet peer dependency "@types/react@>=16".`

### Your environment

*   **OS**: MacOS
*   **Packages**: 
*   **Env**: 

node -v: v13.12.0
npm -v: 6.14.4

### Steps to reproduce

`npm init --yes && npm i react react-dom react-markdown`

### Expected behavior

Maybe the field peerDependenciesMeta can be added to package.json, so the dependency may be made optional.

With this, if the dependency is found on the project, versions will be compared, but no warning will be thrown if not installed.

```
  "peerDependenciesMeta": {
    "@types/react": {
      "optional": true
    }
  },
```

### Actual behavior

react-markdown installed in the non typescript env without peer dependency warning

KieraDOG:
Closes #614

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,11 @@
     "@types/react": ">=16",
     "react": ">=16"
   },
+  "peerDependenciesMeta": {
+    "@types/react": {
+      "optional": true
+    }
+  },
   "devDependencies": {
     "@babel/core": "^7.0.0",
     "@babel/preset-env": "^7.0.0",


ChristianMurphy:
@KieraDOG I'm not able to reproduce the issue https://github.com/remarkjs/react-markdown/issues/614#issuecomment-877186928
I'm also cautious about adding `peerDependenciesMeta` as it is not supported on npm 6 (https://docs.npmjs.com/cli/v6/configuring-npm/package-json) which you specifically mention testing with.
It is only on npm 7+ (https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta) and yarn (https://classic.yarnpkg.com/en/docs/package-json#toc-peerdependenciesmeta)

KieraDOG:
@ChristianMurphy 
> But it would still cause errors for typescript users who could install the module but may not have `@types/react` installed yet.

It should not be the responsibility of this package to check or insure them to install types, right? It's peer dependencies, and the responsibility for it is the package might not work without it, however, react markdown should definitely work without typescript.

ChristianMurphy:
> It's peer dependencies, and the responsibility for it is the package might not work without it

https://github.com/remarkjs/react-markdown/issues/614#issuecomment-877246756

> react markdown should definitely work without typescript

It does

wooorm:
The situation that would result from merging this is bad for TS users. The current situation is OK for non-TS users. As detailed in the conversations of this PR and the related issue, I think the current state is a better trade-off.

Thanks for your work but I don’t think this makes sense at this moment for this project.

belgattitude:
Sorry I haven't seen this P/R and created the same with some more explanation

@wooorm 

> The situation that would result from merging this is bad for TS users

I don't think so: https://github.com/remarkjs/react-markdown/pull/654#issuecomment-957572565. 

tldr; all react/typescripts users have the @types/react installed, isn't it ? Without the optional in peer in react-markdown, the situation is that they're required as deps and not in dev-deps. The optional dependency permits that

ChristianMurphy:
> all react/typescripts users have the @types/react installed, isn't it ?

This is not always the case.

> Without the optional in peer in react-markdown, the situation is that they're required as deps and not in dev-deps. The optional dependency permits that

Using `dependencies` or `devDependencies` already supported without any changes, see https://github.com/remarkjs/react-markdown/pull/654#issuecomment-957694180 for more explanation

Issue #618, opened 2021-07-21
Ability to specify a custom text renderer/component removed in v6.0.0


fastfedora:
### Subject of the issue

Previous versions of `react-markdown` had the ability to provide a custom text renderer (for example, to filter the text or wrap all text in a Material UI `Typography` element). This functionality appears to have been removed in version 6.0.0, with no mention in the changelog or release notes. 

The only way I figured it out was noting there was no mapping for the old `text` render in [this conversion chart](https://github.com/remarkjs/react-markdown/blob/main/changelog.md#600---2021-04-15) and then searching through the code to find that references to the old text renderer were removed.

At a minimum, it would be convenient if the removal of this functionality was documented somewhere (this issue may provide that). But it would also be great to have this functionality back.

### Old behavior

In 5.x and prior, a custom renderer could be provided via the `text` renderer that would apply to all text nodes:

```
  <ReactMarkdown
    renderers={{
      text: ({ children }) => <Typography variant="body1">{children}</Typography>
    }}
    ...
  />
```

### Expected behavior

A mapping would exist for the `components` to provide the same functionality that was available in previous versions, where all text nodes were rendered with the custom renderer:

```
  <ReactMarkdown
    components={{
      text: ({ children }) => <Typography variant="body1">{children}</Typography>
    }}
    ...
  />
```

### Actual behavior

There is no way to override the rendering of text nodes anymore.

Murderlon:
Hi, this has been discussed and answered in this discussion: https://github.com/remarkjs/remark/discussions/700. Let us know if anything is still unclear. 

Closing as this is a duplicate question: #609

--- 

@ChristianMurphy @wooorm perhaps we can update the docs as this seems to be a recurring issue?

fastfedora:
I just found that issue. Apologies for the duplicate issue. Can the CHANGELOG for 6.0.0 perhaps be updated to indicate this though? That would make things a lot easier, since it seems several people are having the same challenge in upgrading.

fastfedora:
Also, discussion 700 proposes that the `p` component/renderer might be able to be used as a replacement. Is there anyone familiar with the codebase that can confirm that every text note is wrapped in a `p`? (I can try doing it, but I'm not familiar with the codebase).

wooorm:
No, not every `text` is in a `p`. And `p`s contain other things (such as emphasis).
You can inspect the syntax tree on [AST Explorer](https://astexplorer.net) (select markdown)

[...]

Issue #609, opened 2021-06-29
Text option in custom components


Fanner487:
I'm trying to upgrade from `v4.3.1` to `v6.0.2`. All is going well with refactoring the `renderers` prop to `components`.

However, I am having issues finding a way to create a custom component on the `text` attribute that was in `v4.3.1`. Reading the [Changelog](https://github.com/remarkjs/react-markdown/blob/main/changelog.md) and the conversion table I see that there is no conversion for the `text` attribute so my code won't get rendered properly. I was wondering if there is a way to do it in the latest version?

| Type (`renderers`) | Tag names (`components`) |
| - | - |
| `blockquote` | `blockquote` |
| `break` | `br` |
...
| `tableCell` | `td`, `th`**‡** |
| `text` | |
| `thematicBreak` | `hr` |

### This is my code:

```js
const components = {
  text: ({node, ...props}) => {
    return props.value.split(' ').map((word, index) => (
      <CustomComponent />
    ));
  },
  inlineCode: (props) => {
    return props.value.split(' ').map((word, index) => (
      <span key={index}>{`${word} `}</span>
    ));
  },
};
```

```js
<ReactMarkdown
  children={formattedSource}
  linkTarget={'_blank'}
  components={components}
/>
```

### Versions
```json
{
  "react": "^17.0.2",
  "react-markdown": "^6.0.2",
}
```

ChristianMurphy:
See the related discussion https://github.com/remarkjs/remark/discussions/700
It sounds like you have a similar use case

Fanner487:
Thank you very much for pointing me in the right direction!

ChristianMurphy:
A concrete example of how this can be supported in the latest version https://codesandbox.io/s/react-markdown-wrap-text-79e70?file=/src/App.js

PR #619, opened 2021-07-21
Add `transformText` prop

fastfedora:
Adds a `transformText` prop to allow text fragments to be transformed (e.g., filtered or wrapped with a React component).

This prop can be used to replace the functionality of the `text` renderer that was removed in v6.0.0. 

This specifically addresses the cross-cutting concern of altering the rendering of text within `react-markdown` that was previously supported by this library. While it is possible to override every potential HTML tag under `components` that might render text, doing this is unnecessarily complex and requires additional configuration for each plugin that adds new potential tags (e.g., `gfm`). The `transformText` prop provides a single place for all text transformations & rendering to occur.

It can be used, for example, to wrap all text within `Typography` elements or filter sensitive information. For example, to ensure all text is rendered with the correct typography, the following can be used:

```
<ReactMarkdown 
  transformText={(text, index) => 
    <Typography key={index} variant="body1" component="span" inline>{text}</Typography>
  }
>
  {markdown}
</ReactMarkdown>
```

Closes GH-618 and GH-609, and addresses the issues raised in the discussion at remarkjs/remark#700

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -124,6 +124,10 @@ render(<ReactMarkdown remarkPlugins={[gfm]} children={markdown} />, document.bod
 *   `transformImageUri` (`(src, alt, title) => string`, default:
     [`./uri-transformer.js`][uri], optional)\
     Same as `transformLinkUri` but for images
+*   `transformText` (`(text, index, node, parentNode) => string | ReactNode`,
+    default: `undefined`, optional)\
+    Applies any transformations to each fragment of text to return either
+    modified text or a React component used to render the text.
 *   `components` (`Object.<string, Component>`, default: `{}`)\
     Object mapping tag names to React components
 *   `remarkPlugins` (`Array.<Plugin>`, default: `[]`)\
--- a/src/ast-to-react.js
+++ b/src/ast-to-react.js
@@ -69,6 +69,13 @@ exports.hastChildrenToReact = childrenToReact
  * @param {string?} title
  * @returns {string}
  *
+ * @callback TransformText
+ * @param {string} text
+ * @param {number} index
+ * @param {Text} node
+ * @param {Element|Root} parentNode
+ * @returns {string|ReactNode}
+ *
  * @callback TransformLinkTarget
  * @param {string} href
  * @param {Array.<Comment|Element|Text>} children
@@ -148,6 +155,7 @@ exports.hastChildrenToReact = childrenToReact
  * @property {boolean} [includeElementIndex=false]
  * @property {false|TransformLink} [transformLinkUri]
  * @property {TransformImage} [transformImageUri]
+ * @property {TransformText} [transformText]
  * @property {string|TransformLinkTarget} [linkTarget]
  * @property {Components} [components]
  */
@@ -182,7 +190,16 @@ function childrenToReact(context, node) {
         !tableElements.has(node.tagName) ||
         child.value !== '\n'
       ) {
-        children.push(child.value)
+        children.push(
+          context.options.transformText
+            ? context.options.transformText(
+                child.value,
+                childIndex,
+                child,
+                node
+              )
+            : child.value
+        )
       }
     }
     // @ts-ignore `raw` nodes are non-standard
--- a/src/react-markdown.js
+++ b/src/react-markdown.js
@@ -173,6 +173,7 @@ ReactMarkdown.propTypes = {
   transformLinkUri: PropTypes.oneOfType([PropTypes.func, PropTypes.bool]),
   linkTarget: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
   transformImageUri: PropTypes.func,
+  transformText: PropTypes.func,
   components: PropTypes.object
 }
 
--- a/test/__snapshots__/react-markdown.test.js.snap
+++ b/test/__snapshots__/react-markdown.test.js.snap
@@ -2123,6 +2123,23 @@ exports[`should handle ordered lists with a start index 1`] = `
 </ol>
 `;
 
+exports[`should handle text with custom text transformer 1`] = `
+<p>
+  <span
+    style={
+      Object {
+        "color": "red",
+      }
+    }
+  >
+    Congratulations! 
+  </span>
+  <strong>
+    You won!
+  </strong>
+</p>
+`;
+
 exports[`should handle tight, unordered lists 1`] = `
 <ul>
   
--- a/test/react-markdown.test.js
+++ b/test/react-markdown.test.js
@@ -151,6 +151,28 @@ test('should handle links with custom uri transformer', () => {
   expect(component.toJSON()).toMatchSnapshot()
 })
 
+test('should handle text with custom text transformer', () => {
+  const input = 'Congratulations! **You won!**'
+  /**
+   * @param {string} text
+   * @param {number} index
+   * @returns {string|ReactNode}
+   */
+  const transform = (text, index) =>
+    text.startsWith('Congratulations') ? (
+      <span key={index} style={{color: 'red'}}>
+        {text}
+      </span>
+    ) : (
+      text
+    )
+  const component = renderer.create(
+    <Markdown children={input} transformText={transform} />
+  )
+
+  expect(component.toJSON()).toMatchSnapshot()
+})
+
 test('should use target attribute for links if specified', () => {
   const input = 'This is [a link](https://espen.codes/) to Espen.Codes.'
   const component = renderer.create(


wooorm:
Thanks for the PR.
I gave it some time to noodle in my head, but I don’t really like it.

My main problems are:
- I don’t think it solves your issue: you’re wrapping text inside code, or even stylesheets and scripts, with this, and that either doesn’t look nice (because what’s supposed to be monospace is now sans-serif), is impractical (turning everything red), or broken (text inside scripts/styles/textareas and such should be styled)
- I can’t imagine it being fast to wrap every text node in a document with a component
- There are good alternatives: just plain old CSS, or specifying `p`, `h1-6`, and `li` components.
- It will probably be abused for things where remark/rehype plugins are better suited.

So I’m leaning somewhat strongly towards wontfixing this, but perhaps @ChristianMurphy is in favor?

ChristianMurphy:
I'd lean against it as well.
Mainly that I anticipate that people will mostly want to target some text constructs, but not all.
For example wanting different typography for strong and emphasis content, which could and likely would be overriden by blanket typography styles applied to all text.

remark and rehype plugins allow for that flexibility of targeting some content types.

fastfedora:
I'm a bit frustrated that this was closed over the weekend without any opportunity to respond, or without addressing the original problem described in the PR description.

Most of the comments listed above don't address the actual problem, nor the fact that this functionality previously existed in this project and served a valuable purpose there.

Specific responses to the comments:

> * I don’t think it solves your issue: you’re wrapping text inside code, or even stylesheets and scripts, with this, and that either doesn’t look nice (because what’s supposed to be monospace is now sans-serif), is impractical (turning everything red), or broken (text inside scripts/styles/textareas and such should be styled)

This is not wrapping text inside code (any more than ALL text within a web app is text wrapped within code). This is rendering text using a specific React component. It helps to implement Material Design as implemented by the Material-UI project, where all text is wrapped in a `Typography` component.

In theory, this would wrap stylesheets & scripts—except that standard Markdown has no way to specify stylesheets or scripts, so this is not an issue. Even if someone added a plugin that allowed stylesheets or scripts to be specified, they could be specifically excluded using the `parentNode` parameter. 

As for solving my issue, it absolutely solves my issue. I already switched to using the PR branch in my code, which fixed things elegantly. It could also solve the issues linked to in the description of this PR.

> * I can’t imagine it being fast to wrap every text node in a document with a component

Why wouldn't it be? You're likely wrapping only a handful of components. I would guess the number of text nodes in most Markdown documents is not more than 50-100. Besides, if you had to pre-process all those nodes, it would take just as long and likely longer.

> * There are good alternatives: just plain old CSS, or specifying `p`, `h1-6`, and `li` components.

No, those are not good alternatives. They are hacks. They break the encapsulation that a component like `Typography` provides. 

Yes, if you are using a framework where you use global CSS, you can specify plain old CSS. But if you are using a component-based design framework like Material UI, then by overriding with plain old CSS, you prevent the theming system from working properly and introduce potential issues with upgrades. 

The point of using a component like `Typography` is that it encapsulates the entire design. Using global CSS to subvert this is a hack.

> * It will probably be abused for things where remark/rehype plugins are better suited.

Doing this with a remark/rehype plugin is a lot more complicated. Before creating this PR, I went through all of the current remark/rehype plugins looking for ones that could help solve this problem, and didn't find any. 

This solution already addresses the desire not to have a separate "text" renderer, like in versions < 6.0 by mimicking the functionality of `transformLinkUri` and `transformImageUri` (which, as far as I understand, could also be implemented in remark/rehype). 

> Mainly that I anticipate that people will mostly want to target some text constructs, but not all.
> For example wanting different typography for strong and emphasis content, which could and likely would be overriden by blanket typography styles applied to all text.

Again, that is not what the point of this PR is. This is to address targeting *all* text constructs. If only specific text constructs need to be targeted, that is better done using the `components` prop, which allows overriding specific elements.

ChristianMurphy:
> without addressing the original problem described in the PR description.

@fastfedora Again, the functionality you want is supported without code changes to `react-markdown`, through a rehype plugin.
Here is a concrete example https://codesandbox.io/s/react-markdown-wrap-text-79e70?file=/src/App.js

> > Doing this with a remark/rehype plugin is a lot more complicated

The plugin is 10 lines of code, with prettier formatting, it is not as complicated as you make it out to be.

> Again, that is not what the point of this PR is. This is to address targeting all text constructs. If only specific text constructs need to be targeted, that is better done using the components prop, which allows overriding specific elements

I'm aware of your goal.
Using a rehype plugin offers the flexibility to support both your use case, and the use case targeting specific constructs, without adding to the API surface of `react-markdown`.
And allowing both use-cases to leverage the `components` prop.

fastfedora:
> > without addressing the original problem described in the PR description.
> 
> @fastfedora Again, the functionality you want is supported without code changes to `react-markdown`, through a rehype plugin.
> Here is a concrete example https://codesandbox.io/s/react-markdown-wrap-text-79e70?file=/src/App.js

Thank you for providing this example. It would have saved a lot of time and trouble if someone could have provided this last month. I will try this in a future update, and if it works, switch back to the main branch of the project.

As feedback to the project management for this project, it would be helpful if concerns were not so quickly dismissed and there was an effort made to address concerns better.

I ran into this problem upgrading this library, where this functionality was removed with no documentation or notice other than a blank cell in the `components` mapping table. It was just not me running into this problem either, as noted in the linked issues above.

I not only tried extensively to explain the problem in issue GH-618, I also did research into whether any existing rehype/remark plugins could fix my problem before creating this PR and submitting it. 

I will admit I did not go so far as to learn how to write a custom rehype/remark plugin, since it was clear my concerns weren't being addressed in GH-618 (as in comments in this PR—talking about all sorts of use cases *but* the use case of wrapping in a `Typography` component), and it was an easy fix to the library itself.

In the future, consider when features are removed, to better document that and, if it's possible, provide an example like this that shows how someone without deep knowledge of the library codebase & its ecosystem can quickly replicate the functionality. 

Since I'm not the only one who has had this issue, I recommend linking to this example somewhere in your documentation, so others know how they can upgrade to newer versions of this library.

ChristianMurphy:
> As feedback to the project management for this project

A friendly reminder, this is open source, not a product.
There is no paid "project management" team, just volunteers.
The issue tracker and discussions are open, if you'd like to help triage issues and answer questions, are are welcome to!

> it would be helpful if concerns were not so quickly dismissed and there was an effort made to address concerns better.
>
> as in comments in this PR—talking about all sorts of use cases but the use case of wrapping in a `Typography` component

The XY problem is fairly common when folks ask questions.
There is no intent to dismiss your concern, the questions from others consistently center around getting a better understanding of the root problem you are trying to solve, while your responses keep centering on the implementation detail that it must be `Typography`, taking a long time to offer context on _why_.
Getting at the root issue that people are trying to solve is important to offering an effective solution.

The question of if every component _should_ be wrapped with `Typography` is still debatable.
The alternatives are still preferable in most cases, and potentially in your case as well.

> But if you are using a component-based design framework like Material UI, then by overriding with plain old CSS, you prevent the theming system from working properly and introduce potential issues with upgrades.
>
> Using CSS to subvert this is a hack.

This is something that should be raised with Material UI.
I'm not sure it's accurate, and if it is, it sounds like a bug with Material UI react.
CSS is how things are styled on the web, using plain CSS should not be a hack.

greg-pendlebury:
> @fastfedora Again, the functionality you want is supported without code changes to `react-markdown`, through a rehype plugin. Here is a concrete example https://codesandbox.io/s/react-markdown-wrap-text-79e70?file=/src/App.js

Thanks so much for this snippet. Has helped me a lot with similar issues. I would note that it is not valid TypeScript because of how `Components` is typed. You are only allowed to use valid HTML tags... I picked `section`... whether it blows up for me later I shall discover.

I don't have our upgrade nailed yet, but this was a good start.

ChristianMurphy:
You can add custom elements in typescript too
https://medium.com/@joelmalone/get-jsx-to-recognise-your-custom-element-in-react-or-preact-bf08d7522208

greg-pendlebury:
> You can add custom elements in typescript too https://medium.com/@joelmalone/get-jsx-to-recognise-your-custom-element-in-react-or-preact-bf08d7522208

Interesting. I did try for a while Yesterday to see how simple that would be, but couldn't get anything working with React. Seems like Preact has a more flexible way of injecting generics. Also having just been burning by hacking around in module loading for another issue (PDF.js has an odd pattern for loading work threads) I would forecast this approach is possibly very fragile for Jest module loading.

In any event, 'section' worked perfectly for me, and our custom parser strips them back out anyway. It won't be portable to everyone's pipeline, but worked fine for me.

ChristianMurphy:
> Interesting. I did try for a while Yesterday to see how simple that would be, but couldn't get anything working with React. Seems like Preact has a more flexible way of injecting generics.

React and Preact use the same typing approach.
Though in react 19 they did change the namesspaces a bit https://stackoverflow.com/questions/79341174/react-19-typescript-add-typeof-web-component

> Also having just been burning by hacking around in module loading for another issue (PDF.js has an odd pattern for loading work threads) I would forecast this approach is possibly very fragile for Jest module loading.

Huh? What does that have to do with either the typing or the custom element that will be re-mapped to a new element before leaving this module?

> In any event, 'section' worked perfectly for me, and our custom parser strips them back out anyway. It won't be portable to everyone's pipeline, but worked fine for me.

Glad to hear it!

PR #623, opened 2021-07-28
Add `VideoComponent` into types.

SalahAdDin:
Adding types for video component.

### DIFF
--- a/src/ast-to-react.js
+++ b/src/ast-to-react.js
@@ -101,6 +101,10 @@ exports.hastChildrenToReact = childrenToReact
  * @param {ReactBaseProps & ReactMarkdownProps & {level: number}} props
  * @returns {ReactNode}
  *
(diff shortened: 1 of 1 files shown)


ChristianMurphy (changes requested):
I don't think this is a direction we want to go trying to match the JSX intrinsic element spec.
Instead reflecting it back may make sense.
Something like modifying `NormalComponents` and `Components`

```ts
export type NormalComponents = {
    [P in keyof IntrinsicElements]: (props: IntrinsicElements[P] & ReactMarkdownProps) => ReactNode
}

export type Components = Partial<Omit<NormalComponents, keyof SpecialComponents> & SpecialComponents>;
```

wooorm:
@SalahAdDin #624 seems to solve your original problem nicely. What do you think?

PR #640, opened 2021-08-28
Allow SpecialComponents to be any ComponentType

Methuselah96:
### Initial checklist

### Description of changes

Changes the `SpecialComponents` to allow them to be non-functional.

My use case is adding styles to the components with Material-UI. Here's an example of what I'm trying to do:

```
export default function Markdown(props: ReactMarkdownOptions) {
  return (
    <ReactMarkdown
      components={{
        li: withStyles(styles)(
          ({
            classes,
            color,
            node,
            ref,
            ...props
          }: JSX.IntrinsicElements['li'] &
            ReactMarkdownProps & {
              checked: boolean | null;
              index: number;
              ordered: boolean;
            } & WithStyles<typeof styles>) => (
            <li className={classes.listItem}>
              <Typography paragraph component="span" {...props} />
            </li>
          ),
        ),
      }}
      remarkPlugins={[remarkGfm]}
      {...props}
    />
  );
}
```

### DIFF
--- a/lib/ast-to-react.js
+++ b/lib/ast-to-react.js
@@ -1,4 +1,9 @@
 /**
+ * @template T
+ * @typedef {import('react').ComponentType<T>} ComponentType<T>
+ */
+
+/**
  * @typedef {import('react').ReactNode} ReactNode
  * @typedef {import('unist').Position} Position
  * @typedef {import('hast').Element} Element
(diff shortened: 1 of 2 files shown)


lib/ast-to-react.js
```diff
@@ -1,3 +1,8 @@
+/**
+ * @template T
+ * @typedef {import('react').ComponentType<T>} ComponentType<T>
+ */
```
Methuselah96, line 4: This needs to be in a separate comment block in order the template to not apply to all the types in the block.

lib/ast-to-react.js
```diff
@@ -44,33 +49,21 @@
...
- * @callback TableRowComponent
- * @param {JSX.IntrinsicElements['tr'] & ReactMarkdownProps & {isHeader: boolean}} props
- * @returns {ReactNode}
+ * @typedef {JSX.IntrinsicElements['code'] & ReactMarkdownProps & {inline?: boolean}} CodeProps
+ * @typedef {JSX.IntrinsicElements['h1'] & ReactMarkdownProps & {level: number}} HeadingProps
+ * @typedef {JSX.IntrinsicElements['li'] & ReactMarkdownProps & {checked: boolean|null, index: number, ordered: boolean}} LiProps
+ * @typedef {JSX.IntrinsicElements['ol'] & ReactMarkdownProps & {depth: number, ordered: true}} OrderedListProps
+ * @typedef {JSX.IntrinsicElements['table'] & ReactMarkdownProps & {style?: Object.<string, unknown>, isHeader: boolean}} TableCellProps
+ * @typedef {JSX.IntrinsicElements['tr'] & ReactMarkdownProps & {isHeader: boolean}} TableRowProps
+ * @typedef {JSX.IntrinsicElements['ul'] & ReactMarkdownProps & {depth: number, ordered: false}} UnorderedListProps
```
ChristianMurphy, line 58: I'm curious on the thinking of creating dedicated prop types vs inlining this in the `ComponentType<>`.
Are these props something you do/plan to import directly?

ChristianMurphy (approved):
Thanks @Methuselah96!

lib/ast-to-react.js
Methuselah96: Yes, they would be very helpful for me to import directly. In my example:
```
export default function Markdown(props: ReactMarkdownOptions) {
  return (
    <ReactMarkdown
      components={{
        li: withStyles(styles)(
          ({
            classes,
            color,
            node,
            ref,
            ...props
          }: JSX.IntrinsicElements['li'] &
            ReactMarkdownProps & {
              checked: boolean | null;
              index: number;
              ordered: boolean;
            } & WithStyles<typeof styles>) => (
            <li className={classes.listItem}>
              <Typography paragraph component="span" {...props} />
            </li>
          ),
        ),
      }}
      {...props}
    />
  );
}
```

becomes:

```
export default function Markdown(props: ReactMarkdownOptions) {
  return (
    <ReactMarkdown
      components={{
        li: withStyles(styles)(
          ({
            classes,
            color,
            node,
            ref,
            ...props
          }: LiProps & WithStyles<typeof styles>) => (
            <li className={classes.listItem}>
              <Typography paragraph component="span" {...props} />
            </li>
          ),
        ),
      }}
      {...props}
    />
  );
}
```

ChristianMurphy: Does `LiProps & WithStyles<typeof styles>` need to be explicitly added?
Shouldn't this be inferred by TSC without an explicit annotation? :thinking:

Methuselah96: Unfortunately I think it does need to be explicit. All TypeScript sees is that the function is getting passed into `withStyles`. I don't think it's smart enough to then try to figure out how to conform to the type for the `components` prop.

ChristianMurphy: That is unfortunate.
Oh well, thanks for the additional context!

wooorm (approved):

wooorm:
Sorry, forgot about this PR. Released!

PR #644, opened 2021-09-02
fix: react-is version

guoyunhe:
Allow react-is dependency version to match react and react-dom version

### Initial checklist

*   [ ] If applicable, I’ve added docs and tests

### Description of changes

TODO

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "comma-separated-tokens": "^2.0.0",
     "prop-types": "^15.0.0",
     "property-information": "^6.0.0",
-    "react-is": "^17.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -84,7 +84,7 @@
     "comma-separated-tokens": "^2.0.0",
     "prop-types": "^15.0.0",
     "property-information": "^6.0.0",
-    "react-is": "^17.0.0",
+    "react-is": ">=16",
```
ChristianMurphy, line 87: :thinking: If we just unlock the version, npm and yarn will still install latest by default, wouldn't they?
I wonder if this should be moved to a peerDependency?

PR #654, opened 2021-11-01
chore(peer-deps): mark @types/react as optional

belgattitude:
### Initial checklist

### Description of changes

`package.json` declares a peerDependency to @types/react. 

this P/R adds the peerDependecyMeta to allow those types to be optional.

Why ?

Without it, we somehow require the @types/react to be installed by the consumer as dependency (npm, yarn will only emit a warning).

But generally @types/* are installed as devDependency (or not at all when js only projects)

Also not that `yarnpkg/doctor` will choke on this (strict check) and emit an error:

image

This P/R should fix the error. there's a reproduction too: https://github.com/belgattitude/nextjs-monorepo-example/pull/730

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,6 +98,11 @@
     "@types/react": ">=16",
     "react": ">=16"
   },
+  "peerDependenciesMeta": {
+    "@types/react": {
+      "optional": true
+    }
+  },
   "devDependencies": {
     "@types/react": "^17.0.0",
     "@types/react-dom": "^17.0.0",


belgattitude:
Why does it complain about the template ? I haven't removed it :smile:

wooorm:
You removed the final part, which read “do note remove: ...” :p

ChristianMurphy:
This looks to be the same as https://github.com/remarkjs/react-markdown/pull/615?

The result from that discussion:

> The situation that would result from merging this is bad for TS users. The current situation is OK for non-TS users. As detailed in the conversations of this PR and the related issue, I think the current state is a better trade-off.
>
> Thanks for your work but I don’t think this makes sense at this moment for this project.

https://github.com/remarkjs/react-markdown/pull/615#issuecomment-877374710 still makes sense today. :thinking: 
The yarn "error" appears to be easily resolved by installing the peer dependency.

belgattitude:
@ChristianMurphy I read the comment and understand the idea. There's no good and right but I would consider it like this:

1. IMHO TS people are used to add @types/* in their devDependencies (not in dependencies). As an example it's common to add @types/react to devDeps as per their [instructions](https://reactjs.org/docs/static-type-checking.html#type-definitions). There's often a dedicated mention in install notes. And it's easy cause if you forget them, typescript will scream on you early enough :)
2. With the current peerDependencies setup (without optional), we can be confident that ts people using react-markdown, have the types installed (as deps or devDeps). Especially since most typescript/react users will already have '@types/react'.  Who can work in react/ts without them ?  (to say there's no bc here)
3. So the question to adding optional is a no-op in my opinion. It's just allowing us to set the @types/react in devDependencies rather than forcing us to put them in dependencies to strictly conform to what the package manager (yarn, npm...) expects. And also we're inline with the react recommended install (put @types/react in devDeps).

Would it make more sense ?

ChristianMurphy:
Some notes on the behavior of different package managers.

## npm <7

npm does not automatically install `peerDependencies`, instead warning.
That warning goes away if `react` and `@types/react` are added to `dependencies` or `devDependencies`

## npm >=7

npm automatically installs `peerDependencies`.
The neither `react` or `@types/react` needs to be added to package.json
If the version of `@types/react` needs to be customized, it can be done through either `dependencies` or `devDependencies`

## yarn 1

yarn does not automatically install `peerDependencies`, instead warning.
That warning goes away if `react` and `@types/react` are added to `dependencies` or `devDependencies`

---

> It's just allowing us to set the @types/react in devDependencies rather than forcing us to put them in dependencies to strictly conform to what the package manager

This is already supported today, without any changes.

belgattitude:
Great summary, I didn't the changes. Even realized that npm v2 & 3 were installing peer by default?

I just want to emphasize that in this P/R peerDependencies remains unchanged, the change is `peerDependenciesMeta.optional: true`. 

Following your notes, that change would uniformize the behaviour (yarn 1&2&3, npm 6&7, pnpm ?...) which looks a good move to me. Except if you're looking to auto-install for npm v7 users ? 

I know it looks a detail painted by package managers weirdness, but as I see this: go peerMeta optional :smile: 

Basically: 

- @types/react are a peer, and they're optional (they won't be auto installed, maybe show a warning in a consistent way)
- optional won't make a hard time for typescript users, it's @types/react (without them a ts project just don't work)  
- and it will fix some trouble (notably the one in the desc :)

What's your thoughts ?

ChristianMurphy:
It comes back to _**why**_ make `@types/react` optional.

I'm familiar with the reasons why to keep `@types/react` _required_, described in https://github.com/remarkjs/react-markdown/pull/615
But summarized as: unified currently aims to support both JS and TS, for TS users there is an expectation that types `dependencies` or `peerDependencies` (a recent example https://github.com/syntax-tree/unist-util-select/issues/12), so it is either automatically installed or there is a clear message it is required.
Having the `peerDependencies` required benefits TypeScript users, and does not harm JavaScript users.

The reason provided to make `peerDependencies`  _optional_ in https://github.com/remarkjs/react-markdown/pull/654#issue-1041366725 and https://github.com/remarkjs/react-markdown/pull/615#issuecomment-957594052
Was that: required `peerDependencies` need to be installed as `dependencies` (not `devDependencies`) by adopters.
But, as highlighted in https://github.com/remarkjs/react-markdown/pull/654#issuecomment-957694180, this isn't strictly the case, _required_ `peerDependencies` can also be installed as `devDependencies` without warnings or errors.

Leaving the question: why make the `peerDependency` optional? with the answer unclear.

belgattitude:
> why make the peerDependency optional? with the answer unclear.

I mean it's a choice... 'peerDevDependencies' does not exists, and peerDependencies has different interpretations with npm v7. js users probably don't care. ts users neither in my understanding. Outcome of using optional:true... seems we'll have a consistent way of install, a warning that probably disappear... and maybe other things good or bad.

So it's not big deal (for me). I just opened a P/R cause I faced an issue with a specific tool (yarn doctor) and saw warnings, since then I moved my @types/react to deps and while it's not beautiful it does the job.

Decision of "optional: true" is really up to you, feel free to close I won't get mad at this :smile:

ChristianMurphy:
Trying out

_package.json_

```json
{
  "name": "yarn-doctor-peer-test",
  "version": "1.0.0",
  "description": "",
  "dependencies": {
    "react": "^17.0.0",
    "react-dom": "^17.0.0",
    "react-markdown": "^6.0.0"
  },
  "devDependencies": {
    "@types/react": "^17.0.0"
  }
}
```

with yarn 2, running `yarn install` no warnings are flagged.

---

> specific tool (yarn doctor) and saw warnings

yarn doctor is specifically for package maintainers

> A cli tool to help package maintainers support pnp

\- https://www.npmjs.com/package/@yarnpkg/doctor

If you are publishing your code to the npm/yarn registry, then yes the `peerDependency` rule applies, missing the dependency would break downstream adopters.
As a non-published package, many of these rules don't apply.

Running `yarn dlx @yarnpkg/doctor . ` on `react-markdown` itself, the peerDependency issue is not flagged because the dependency is noted correctly.

belgattitude:
So I'll close. THanks for the detailed debugging.  I'll open something in yarn doctor as it seems to belong there. 



BTW I don't use it for pnp actually just for monorepo integrity (boundaries) checks and be sure my packages deps are clearly stated.

PR #667, opened 2021-12-31
fix: support custom component type from a plugin

Divlo:
### Initial checklist

### Description of changes

This PR fixes a bug: currently using a `rehypePlugin`, that replaces in the string (`props.children`), all the occurrences that apply a regex (for example the regex `/d/` as you can see in the unit tests) by a custom React component defined in `props.components`, it doesn't replace, it removes the content.

Basic usage:

```jsx
    <Markdown
      remarkPlugins={[gfm]}
      // the plugin that replaces the letter 'd' in the string 'abcdef' by the 'customType` component
      rehypePlugins={[pluginCustomComponentType]}
      components={{
         customType: (props) => {
  			return <div>{props.value}-custom</div>
		}
      }}
    >
      {'abcdef'}
    </Markdown>
```

Currently, it outputs: `<p>abcef</p>`, with the PR, it will output: `<p>abc<div>d-custom</div>ef</p>`.
The example is quite simple to stay minimal.

Here is an example of a real-world use case: replace all the `:emoji:` in a string, to use the Emoji react component from [emoji-mart](https://www.npmjs.com/package/emoji-mart), (`:emoji:` being `:wave:`, `:+1:`, etc.).

**Note:** This bug is actually a regression, in `react-markdown@5`, it has the right behavior, but not with the latest released version v7.1.1, it started breaking since v6 apparently.

Friendly ping @ChristianMurphy and @wooorm. :smile:

### DIFF
--- a/lib/ast-to-react.js
+++ b/lib/ast-to-react.js
@@ -141,6 +141,20 @@ export function childrenToReact(context, node) {
     } else if (child.type === 'raw' && !context.options.skipHtml) {
       // Default behavior is to show (encoded) HTML.
       children.push(child.value)
+    } else if (!context.options.skipHtml && context.options.components) {
+      const {type, ...props} = child
+      /** @type {string} */
+      const component =
+        context.options.components && own.call(context.options.components, type)
+          ? context.options.components[type]
+          : type
+      const hasCustomComponentType = Object.keys(
+        context.options.components
+      ).includes(type)
+      if (hasCustomComponentType) {
+        const key = [type, childIndex].join('-')
+        children.push(React.createElement(component, {...props, key}))
+      }
     }
   }
 
--- a/test/test.jsx
+++ b/test/test.jsx
@@ -1406,4 +1406,114 @@ test('should crash on a plugin replacing `root`', () => {
   }, /Expected a `root` node/)
 })
 
+/**
+ * @param {string} string
+ * @param {number} start
+ * @param {number} end
+ * @returns {import('unist').Literal<string>}
+ */
+const extractText = (string, start, end) => {
+  const startLine = string.slice(0, start).split('\n')
+  const endLine = string.slice(0, end).split('\n')
+  return {
+    type: 'text',
+    value: string.slice(start, end),
+    position: {
+      start: {
+        line: startLine.length,
+        column: startLine[startLine.length - 1].length + 1
+      },
+      end: {
+        line: endLine.length,
+        column: endLine[endLine.length - 1].length + 1
+      }
+    }
+  }
+}
+
+/** @type {import('unified').Plugin<Array<void>, import('unist').Literal<string>>} */
+const pluginCustomComponentType = () => {
+  const dRegex = /d/
+  /** @type {import('unified').Transformer<import('unist').Literal<string>>} */
+  const transformer = (tree) => {
+    visit(
+      /** @type {import('unist').Literal<string>} */ tree,
+      /** {@type string} */ 'text',
+      (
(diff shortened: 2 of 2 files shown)


wooorm:
I don‘t understand why you want an emoji picker in your content. You can use https://github.com/remarkjs/remark-gemoji or https://github.com/search?q=rehype+emoji fine?

Also, you can create a standard hast element. You shouldn’t need to create custom nodes in hast.

lib/ast-to-react.js
```diff
@@ -141,6 +141,20 @@ export function childrenToReact(context, node) {
...
+        context.options.components && own.call(context.options.components, type)
+          ? context.options.components[type]
+          : type
+      const hasCustomComponentType = Object.keys(
+        context.options.components
+      ).includes(type)
+      if (hasCustomComponentType) {
+        const key = [type, childIndex].join('-')
+        children.push(React.createElement(component, {...props, key}))
+      }
```
ChristianMurphy, line 157: components generally refer to an element type.
I think this would muddle/confuse that by being either: an element type or a node type.

test/test.jsx
```diff
@@ -1406,4 +1406,114 @@ test('should crash on a plugin replacing `root`', () => {
...
+        const definition = []
+        let lastIndex = 0
+        const match = dRegex.exec(node.value)
+        if (match !== null) {
+          const value = match[0]
+          if (match.index !== lastIndex) {
+            definition.push(extractText(node.value, lastIndex, match.index))
+          }
+
+          definition.push({type: 'customType', value})
```
ChristianMurphy, line 1462: why not leverage a custom element type? (which should be supported today, without any changes)
```suggestion
          definition.push({type: 'element', tagName: 'custom-type', properties: { value }})
```

ChristianMurphy (commented):
I'd second @wooorm's questions.

test/test.jsx
Divlo: It seems like, it is not working, by doing this I've got this browser error message:
```js
ast-to-react.js?002a:121 Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at childrenToReact (ast-to-react.js?002a:121)
    at toReact (ast-to-react.js?002a:185)
    at childrenToReact (ast-to-react.js?002a:125)
    at toReact (ast-to-react.js?002a:185)
    at childrenToReact (ast-to-react.js?002a:125)
    at ReactMarkdown (react-markdown.js?2a9f:114)
```

In case you are asking, I cleared the cache, uninstalled/reinstalled `node_modules`, it is using the latest version of `react-markdown`).

Indeed, if we could do something like that, it would solve the problem. :+1:

Divlo:
Hey! :wave:
Thanks for your fast replies.

> I don‘t understand why you want an emoji picker in your content.

`emoji-mart` is not only an emoji picker, but also an [Emoji](https://github.com/missive/emoji-mart#emoji) component.

> You can use remarkjs/remark-gemoji or github.com/search?q=rehype+emoji fine?

No, it is not what I want to do, I don't want to show the emoji as simple text, as the appearance depends on the machine, I want to use the Twitter emoji look, using the [Emoji](https://github.com/missive/emoji-mart#emoji) react component.

test/test.jsx
ChristianMurphy: https://codesandbox.io/s/custom-element-example-r4mt5

Divlo: Thank you very much, it works!
I will close the PR as no changes are needed.

I could open an issue before opening a PR, my mistake.

PR #692, opened 2022-06-05
Add clarification of `alt` on `img` in docs

cballenar:
### Initial checklist

### Description of changes

Included `alt` as one of the attributes expected in the `img` component as this component outputs both `alt` AND `title`. Not having both can be misleading specially since `alt` is what's most often/easily used when creating an image in markdown `[!alt](src "title")`. The use of alt over title comes not only from practicality but also due to accessibility. Additionally, removed parenthesis as it read better now that there wasn't just one item.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -571,7 +571,8 @@ Other remark or rehype plugins that add support for new constructs will also
 work with `react-markdown`.
 
 The props that are passed are what you probably would expect: an `a` (link) will
-get `href` (and `title`) props, and `img` (image) an `src`, `alt` and `title`, etc.
+get `href` (and `title`) props, and `img` (image) an `src`, `alt` and `title`,
+etc.
 There are some extra props passed.
 
 *   `code`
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -571,7 +571,7 @@ Other remark or rehype plugins that add support for new constructs will also
 work with `react-markdown`.
 
 The props that are passed are what you probably would expect: an `a` (link) will
-get `href` (and `title`) props, and `img` (image) an `src` (and `title`), etc.
+get `href` (and `title`) props, and `img` (image) an `src`, `alt` and `title`, etc.
```
wooorm, line 574:
```suggestion
get `href` (and `title`) props, and `img` (image) an `src`, `alt` and `title`,
etc.
```

(this suggestion adds a line break)

wooorm (approved):

wooorm:
Thank you!

PR #697, opened 2022-07-03
Fix error

john-montgomery2003:
### Initial checklist

*   [ x] I read the support docs 
*   [x ] I read the contributing guide 
*   [x ] I agree to follow the code of conduct 
*   [x ] I searched issues and couldn’t find anything (or linked relevant results below) 
*   [ x] If applicable, I’ve added docs and tests

### Description of changes

This caused an error on our project, the standard defines it as being `catch (e)` this should fix.
Thanks

### DIFF
--- a/lib/ast-to-react.js
+++ b/lib/ast-to-react.js
@@ -406,7 +406,7 @@ function parseStyle(value) {
 
   try {
     style(value, iterator)
-  } catch {
+  } catch (e) {
     // Silent.
   }
 


john-montgomery2003:
its a tiny fix and therefore i have not really added any docs or anything, theres really not much to say, just a small bug fix

ChristianMurphy:
The standard says defining the exception identifer is optional, and the code currently in place is valid.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch

This is not a bug, it is likely your build tool is misconfigured.

Issue #635, opened 2021-08-17
Error importing component in Jest tests with TypeScript


tubbo:
### Initial checklist

- [X] I read the [support docs](https://github.com/remarkjs/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/remarkjs/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/remarkjs/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Aremarkjs&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

react-markdown@npm:7.0.0 [bad77] (via npm:^7.0.0 [bad77])

### Link to runnable example

https://github.com/tubbo/react-markdown-repro

### Steps to reproduce

Install `react-markdown` on a TypeScript project and try to use it in a test. I used CRA in my example but that isn't strictly necessary, as I came across the issue on an Electron project. Trying to configure `transformIgnorePatterns` to whitelist this module also proves impossible, as you have to also whitelist every dependency of the module which is a nightmare.

### Expected behavior

Test should run without having trouble importing the file

### Actual behavior

```
 FAIL  src/community/community.test.tsx
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /[...]/node_modules/react-markdown/index.js:5
    import { ReactMarkdown } from './lib/react-markdown.js';
    ^^^^^^

    SyntaxError: Cannot use import statement outside a module

[...]

PR #702, opened 2022-07-29
Make minified source available to all tools

david-bezero:
### Initial checklist

*   [x] If applicable, I’ve added docs and tests (not applicable)

### Description of changes

Support other tools which cannot work with ES Modules (e.g. Jest). The necessary code is already available but using a proprietary key in package.json (for unpkg). This adds standards-compliant config to make it available to other tooling which needs it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,10 @@
   "main": "index.js",
   "types": "index.d.ts",
   "unpkg": "react-markdown.min.js",
+  "exports": {
+    "require": "./react-markdown.min.js",
+    "default": "./index.js"
+  },
   "files": [
     "lib/",
     "index.d.ts",


david-bezero:
the test failure seems to be because some build tooling is incompatible with the latest change to (experimental) loaders in Node 18.6 (unrelated to this change)

ChristianMurphy:
There is a dual package hazard when exporting both CJS and ESM.
It has been discussed at some length at https://github.com/remarkjs/remark/issues/969#issuecomment-1068875448 and https://github.com/unifiedjs/unified/issues/121#issuecomment-804711549
It can cause real problems, trying to work around Jest-specific limitations.

There are some suggestions on how to make Jest play nicely with ESM in https://github.com/remarkjs/react-markdown/issues/635

david-bezero:
@ChristianMurphy I guess note that the dual package hazard only applies to packages which contain meaningful global state (which, as far as I can see, this one doesn't). Also you're already _providing_ the dual packages; this PR just makes them both _available_ to the tooling, so I hope you might reconsider this decision.

For now, the only workaround available to users is this rather nasty config:

```js
  moduleNameMapper: {
    'react-markdown': '<rootDir>/node_modules/react-markdown/react-markdown.min.js',
  },
```

ChristianMurphy:
The dual package hazard has several pitfalls and risks, global state is just one of them.
In addition, some remark/rehype plugins do use global state.

Having to configure around a bug/limitation of Jest is an acceptable trade-off, vs the risks and known issues of dual packaging.
And I'm hopeful in the future Jest will make ESM fully supported and seamless https://github.com/facebook/jest/issues/9430

PR #736, opened 2023-04-05
Improve README syntax highlight example

dlqqq:
### Initial checklist

### Description of changes

Improves README syntax highlighting example. The current example produces a TypeScript error by default. See here: https://github.com/react-syntax-highlighter/react-syntax-highlighter/issues/479#issuecomment-1267772279

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -354,14 +354,14 @@ ReactDom.render(
         const match = /language-(\w+)/.exec(className || '')
         return !inline && match ? (
           <SyntaxHighlighter
+            {...props}
             children={String(children).replace(/\n$/, '')}
             style={dark}
             language={match[1]}
             PreTag="div"
-            {...props}
           />
         ) : (
-          <code className={className} {...props}>
+          <code {...props} className={className}>
             {children}
           </code>
         )


readme.md
```diff
@@ -354,14 +354,14 @@ ReactDom.render(
         const match = /language-(\w+)/.exec(className || '')
         return !inline && match ? (
           <SyntaxHighlighter
+            {...props}
```
remcohaszing, line 357: It might be better to not even pass spread props at all.

Anyway, I'm not necessarily against it.

readme.md
```diff
@@ -354,14 +354,14 @@ ReactDom.render(
...
             children={String(children).replace(/\n$/, '')}
             style={dark}
             language={match[1]}
             PreTag="div"
-            {...props}
           />
         ) : (
-          <code className={className} {...props}>
+          <code {...props} className={className}>
```
remcohaszing, line 364: The order doesn't matter here.

remcohaszing (approved):

ChristianMurphy (approved):
Thanks @dlqqq!
I'd second @remcohaszing's suggestions.

readme.md
wooorm: `true`, but its not bad to be consistent with the other example!

readme.md
wooorm: I think there was something about `SyntaxHighlighter` passing all other props through on the `div` or so:

> spread props pass arbitrary props to pre tag wrapping code.
> — https://github.com/react-syntax-highlighter/react-syntax-highlighter/tree/efc3f7b7537d1729193b7a472067bcbe6cbecaf1#props

Could be removed too, okay with me. But that can be a different PR!

wooorm (approved):

PR #737, opened 2023-04-06
fix: Optional catch binding is a new feature introduced in ECMAScript…

QiRaining:
… 2019 (ES10), and older versions of browsers might not support it.

image

### Initial checklist

*   [ ] I read the support docs 
*   [ ] I read the contributing guide 
*   [ ] I agree to follow the code of conduct 
*   [ ] I searched issues and couldn’t find anything (or linked relevant results below) 
*   [ ] If applicable, I’ve added docs and tests

### Description of changes

TODO

### DIFF
--- a/lib/ast-to-react.js
+++ b/lib/ast-to-react.js
@@ -413,7 +413,7 @@ function parseStyle(value) {
 
   try {
     style(value, iterator)
-  } catch {
+  } catch (err) {
     // Silent.
   }
 


remcohaszing:
This is a duplicate of https://github.com/remarkjs/react-markdown/pull/697. Optional catch binding is supported as per the [Compatibility](https://github.com/remarkjs/react-markdown#compatibility) section in the readme. You can use Babel to transpile dependencies to support older engines.

remcohaszing:
Duplicate of #697

QiRaining:
> This is a duplicate of #697. Optional catch binding is supported as per the [Compatibility](https://github.com/remarkjs/react-markdown#compatibility) section in the readme. You can use Babel to transpile dependencies to support older engines.

So this tool suggests using Babel or older engines, what was the original intention behind creating this tool?  
Funny suggestions

ChristianMurphy:
This `react-markdown` converts markdown into react.
And targets modern evergreen browsers by default.
It can also be used with legacy browsers, using babel to transpired and polyfill.

The modern browser first approach is done because we don't want to force millions of people to download legacy syntax and polyfills they don't need.

In the cases legacy support is needed, Babel or SWC can we quickly and efficiently add only the syntax and polyfills that you and your specific browser matrix need.

Issue #749, opened 2023-07-05
Add examples regarding new lines / line breaks


stephanbogner:
### Initial checklist

- [X] I read the [support docs](https://github.com/remarkjs/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/remarkjs/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/remarkjs/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Aremarkjs&type=Issues) and couldn’t find anything (or linked relevant results below)

---

### Problem

I am struggling to work with new lines / line breaks and [many others did in the past](https://github.com/remarkjs/react-markdown/issues?q=is%3Aissue+line+break+is%3Aclosed).

**From what I've seen there are three problems:**
1. Lack of knowledge about Markdown itself (e.g. double space before `\n`)
2. Missing knowledge about Markdown flavors / plugins (e.g. [remark-breaks](https://github.com/remarkjs/remark-breaks) to turn soft line endings into hard breaks)
3. Problems with using the library itself

---

#### Regarding 1) and 2)
Could be solved/enhanced by mentioning it next to an appropriate example. Something like "Tip: Read up on [how Markdown treats new lines](https://github.com/remarkjs/react-markdown/issues/240) and have a look at plugins such as [`remark-breaks`](https://github.com/remarkjs/remark-breaks) that can change this behaviour."

#### Regarding 3)
I compiled a few ways to use markdown that I tried from which some failed, which I think also cause some of the confusion.

##### ❌ Case 1
Doesn't work as expected (treats everything as the headline)
```
<ReactMarkdown>
    # I am headline 1

    The **bold brown fox** jumps over the _italic dog_
</ReactMarkdown>
```

HTML:
```
<h1>I am headline 1 The <strong>bold brown fox</strong> jumps over the <em>italic dog</em></h1>
```

##### ✅ Case 2
Works as expected, but has a pitfall:
If a line in the markdown is indented with spaces it (rightfully) gets interpreted as a `pre`-tag.
```
const markdown = `# I am headline 1
    
The **bold brown fox** jumps over the _italic dog_`

[...]

<ReactMarkdown children={markdown}/>
```

HTML:
```
<h1>I am headline 1</h1>
<p>The <strong>bold brown fox</strong> jumps over the <em>italic dog</em></p>
```

##### ✅ Case 3
Works as expected
```
[...]

PR #754, opened 2023-07-14
Changed passing children by props to nest between tags in docs

goofynugtz:
### Initial checklist

### Description of changes
Some of the examples mentioned in the readme are passing children as props which gives a ESLint error of `react/no-children-prop` on Next.js 13 with ESLint configured
```javascript
<ReactMarkdown children={markdown}  />
```
```
yarn run v1.22.19
$ next build

Failed to compile.

./pages/blogs/[slug].tsx
72:13  Error: Do not pass children as props. Instead, nest children between the opening and closing tags.  react/no-children-prop

info  - Need to disable some ESLint rules? Learn more here: https://nextjs.org/docs/basic-features/eslint#disabling-rules
error Command failed with exit code 1.
```
Updated the examples to nest child component between the opening and closing tags.
```javascript
<ReactMarkdown>{markdown}</ReactMarkdown>
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -143,7 +143,7 @@ import remarkGfm from 'remark-gfm'
 const markdown = `Just a link: https://reactjs.com.`
 
 ReactDom.render(
-  <ReactMarkdown remarkPlugins={[remarkGfm]} >
+  <ReactMarkdown remarkPlugins={[remarkGfm]}>
     {markdown}
   </ReactMarkdown>,
   document.body
@@ -251,7 +251,7 @@ A table:
 `
 
 ReactDom.render(
-  <ReactMarkdown remarkPlugins={[remarkGfm]} >{markdown}</ReactMarkdown>,
+  <ReactMarkdown remarkPlugins={[remarkGfm]}>{markdown}</ReactMarkdown>,
   document.body
 )
 ```
@@ -369,7 +369,7 @@ ReactDom.render(
       }
     }}
   >
-  {markdown}
+    {markdown}
   </ReactMarkdown>,
   document.body
 )
@@ -408,7 +408,9 @@ ReactDom.render(
   <ReactMarkdown
     remarkPlugins={[remarkMath]}
     rehypePlugins={[rehypeKatex]}
-  >{`The lift coefficient ($C_L$) is a dimensionless coefficient.`}</ReactMarkdown>,
+  >
+    {`The lift coefficient ($C_L$) is a dimensionless coefficient.`}
+  </ReactMarkdown>,
   document.body
 )
 ```
@@ -526,7 +528,7 @@ Some *emphasis* and <strong>strong</strong>!
 </div>`
 
 ReactDom.render(
-  <ReactMarkdown rehypePlugins={[rehypeRaw]} >{input}</ReactMarkdown>,
+  <ReactMarkdown rehypePlugins={[rehypeRaw]}>{input}</ReactMarkdown>,
   document.body
 )
 ```
--- a/readme.md
+++ b/readme.md
@@ -408,7 +408,7 @@ ReactDom.render(
   <ReactMarkdown
     remarkPlugins={[remarkMath]}
     rehypePlugins={[rehypeKatex]}
-  >{`The lift coefficient ($C_L$) is a dimensionless coefficient.`}</>,
+  >{`The lift coefficient ($C_L$) is a dimensionless coefficient.`}</ReactMarkdown>,
   document.body
 )
 ```
--- a/readme.md
+++ b/readme.md
@@ -143,7 +143,9 @@ import remarkGfm from 'remark-gfm'
 const markdown = `Just a link: https://reactjs.com.`
 
 ReactDom.render(
-  <ReactMarkdown children={markdown} remarkPlugins={[remarkGfm]} />,
+  <ReactMarkdown remarkPlugins={[remarkGfm]} >
+    {markdown}
+  </ReactMarkdown>,
   document.body
 )
 ```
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -143,7 +143,9 @@ import remarkGfm from 'remark-gfm'
 const markdown = `Just a link: https://reactjs.com.`
 
 ReactDom.render(
-  <ReactMarkdown children={markdown} remarkPlugins={[remarkGfm]} />,
+  <ReactMarkdown remarkPlugins={[remarkGfm]} >
```
remcohaszing, line 146:
```suggestion
  <ReactMarkdown remarkPlugins={[remarkGfm]}>
```

readme.md
```diff
@@ -249,7 +251,7 @@ A table:
 `
 
 ReactDom.render(
-  <ReactMarkdown children={markdown} remarkPlugins={[remarkGfm]} />,
+  <ReactMarkdown remarkPlugins={[remarkGfm]} >{markdown}</ReactMarkdown>,
```
remcohaszing, line 254:
```suggestion
  <ReactMarkdown remarkPlugins={[remarkGfm]}>{markdown}</ReactMarkdown>,
```

readme.md
```diff
@@ -403,10 +406,9 @@ import 'katex/dist/katex.min.css' // `rehype-katex` does not import the CSS for
 
 ReactDom.render(
   <ReactMarkdown
-    children={`The lift coefficient ($C_L$) is a dimensionless coefficient.`}
     remarkPlugins={[remarkMath]}
     rehypePlugins={[rehypeKatex]}
-  />,
+  >{`The lift coefficient ($C_L$) is a dimensionless coefficient.`}</ReactMarkdown>,
```
remcohaszing, line 411:
```suggestion
  >
    {`The lift coefficient ($C_L$) is a dimensionless coefficient.`}
  </ReactMarkdown>,
```

readme.md
```diff
@@ -367,7 +368,9 @@ ReactDom.render(
         )
       }
     }}
-  />,
+  >
+  {markdown}
```
remcohaszing, line 372:
```suggestion
    {markdown}
```

readme.md
```diff
@@ -524,7 +526,7 @@ Some *emphasis* and <strong>strong</strong>!
 </div>`
 
 ReactDom.render(
-  <ReactMarkdown rehypePlugins={[rehypeRaw]} children={input} />,
+  <ReactMarkdown rehypePlugins={[rehypeRaw]} >{input}</ReactMarkdown>,
```
remcohaszing, line 529:
```suggestion
  <ReactMarkdown rehypePlugins={[rehypeRaw]}>{input}</ReactMarkdown>,
```

remcohaszing (changes requested):
Thanks! I agree passing children as actual children instead of a prop is better. I just have some nits about spacing.

readme.md
JounQin: This can be plain text without template string?

```jsx
  <ReactMarkdown
    remarkPlugins={[remarkMath]}
    rehypePlugins={[rehypeKatex]}
  >
    The lift coefficient ($C_L$) is a dimensionless coefficient.
  </ReactMarkdown>
```

remcohaszing: It could be, but the spacing rules for JSX often lead to confusion. By this I don’t mean to say we shouldn’t use that, just that there is an argument to make for not doing it.

goofynugtz:
Thanks for the reviews. I have made the suggested changes.

remcohaszing (approved):

ChristianMurphy (commented):
I'm cautious about this.
Yes as a general react practice, using nesting over the `children` prop is ideal.
That said, JSX spacing rules and template string spacing rules inside JSX are a frequent source of confusion for new adopters https://github.com/remarkjs/react-markdown/issues?q=is%3Aissue+JSX+is%3Aclosed

I'd lean against making this change.
I think it will cause more confusion than it clarifies.
But will not block the PR if other maintainers feel strongly otherwise.

remcohaszing:
If we don’t want this, I think we should rename the prop to something else, e.g. `content` or `markdown`. (A semver major is upcoming anyway.)

ChristianMurphy:
> If we don’t want this, I think we should rename the prop to something else, e.g. `content` or `markdown`. (A semver major is upcoming anyway.)


I don't think it necessitates a major rework of the API.
There is a lot of overlap in adopters who don't understand that the two forms of children in react are equivalent and those don't understand the rules of JSX, and vice versa.

We could leave the docs as-is and let adopters choose which syntax they prefer.

---

To per-emptively answer the question: "could we document the rules of JSX next to the example?"
Yes we could, but it wouldn't help much, no matter how many ":warning: IMPORTANT READ THIS!" we put around the example, many newcomers will ignore it, copy the example, and open an issue asking why it doesn't work as they'd expect (having not read the JSX spacing rules).

wooorm:
Did y’all see https://github.com/remarkjs/react-markdown/issues/749#issuecomment-1636768884?

I think people are going to try to pass children whether it works or not. As “text” inside the component, or as a sole expression inside the component. The main similar project to this, `markdown-to-jsx`, uses `children` too. And so do several other similar projects. Previously we used a different field, but a couple majors back, we moved to `children`, because it’s a common expectation from users.

Theoretically, we could throw on that. Or we could ignore it and support another field. I don’t prefer those over the current, sole, `children`.

The differences between `<ReactMarkdown children={x} />`, `<ReactMarkdown>{x}</ReactMarkdown>`, and `<ReactMarkdown># literal X</ReactMarkdown>`, is a bit complex for (somewhat) newcomers to React. But I feel that’s more a React fact. That it’s their responsibility. Which is fine to document with an example appendix *and* improved examples as in GH-749.

I also mostly agree with Christian! (only point that I think we can use *one* style in the examples, so people copy the nicest one, where currently 2 are shown)

wooorm:
Used one style in the docs: `<Markdown>{markdown}</Markdown>`, so that solves this!

Issue #413, opened 2020-04-13
Important security consideration for using target="_blank"


stackola:
Using `target="_blank"` on hyperlinks without using `rel="noopener"` is a problem, because the opened page can access and modify some properties on the opening page.

You can find out more information here:

https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

# Proposed fix

Automatically add `rel="noopener noreferrer"` to all links that are targeting `_blank`, or make `rel` user-accessible like `target`

fdidron:
Hello, I created PR #418 for this, I went with adding a `linkRel` prop instead of automagically adding `rel="noopener noreferrer"`.

Needed this for https://configure.ergodox-ez.com/ :)

fdidron:
Doh there was already an opened PR #350

ChristianMurphy:
This can also be added through a plugin using https://github.com/remarkjs/remark-external-links

wchargin:
Also in the same boat; I actually tried the workaround suggested by
@ChristianMurphy, but wasn’t able to get it to work (found this thread
by searching for `remark-external-links`). Cf. #188.

tomchen:
Very important security fixed, hope either #350 or #418 can be merged very soon.

Actually when `linkTarget` is set to `_blank`, `linkRel` should default to `noopener noreferrer` instead of the initial default `undefined`, although it would complicate the settings a little bit.

wooorm:
The suggested solution is the plugin mentioned above; there will be a release soon which fixes support for plugins

Issue #12, opened 2016-01-19
Option for Setting target="_blank" on Links


Enijar:
Is it feasible to allow this as an option for this package?

Something like the following:

```jsx
<ReactMarkdown source={input} linkTargets="_blank"/>
```

rexxars:
A bit unsure. It's easy enough to add, but I don't know if the API makes sense. I think there are two options:

* Allowing users to pass their own handlers for types, eg `Link: MyCustomLinkComponent`. This would be very flexible and is something I've been wanting to add anyway.
* Adding a `linkTarget` prop that can either be a string or a function. The functions could determine the target based on the node, for instance whether it is a relative URL or not.

I'm leaning towards the first, as that will allow all sorts of flexibility to be added without bloating the renderer with options.

Enijar:
I agree. The first option you listed sounds the best way to go and would open up the API to be much more flexible.

Do you have any inkling as to when such a feature could be ready?

rexxars:
I don't think it should be too much work, so it's more a question of when I'll have the time to prioritize it. I'm hoping it'll be this week, at least.

Enijar:
That would be perfect. I really like this package as I don't have to use the `dangerouslySetInnerHTML` property!

Keep up the good work on this package :+1:

petetnt:
Wouldn't the most optimal case be that [any valid attribute](https://facebook.github.io/react/docs/tags-and-attributes.html) would be valid and applied? When rendering the container, you could just spread the props with `...this.props` (and handling special cases such like you do now).

rexxars:
I've implemented custom renderers in `commonmark-react-renderer` (which powers this component). I'll see if I can get a new release of `react-markdown` out quite soon. When that is out it should be as as easy as:
```js
<ReactMarkdown
  source="Click [here](https://espen.codes/) to visit external site"
  renderers={{Link: props => <a href={props.href} target="_blank">{props.children</a>}}
/>
```

rexxars:
Version 4 is now out, and the above example should work for this use case.

cyberwombat:
I wonder how to do a paragraph like this? I end up `Each child in an array or iterator should have a unique "key" prop. Check the render method of 'Paragraph'`

elimisteve:
In case what I found helps others:

I needed something similar within an Electron app and went with [this solution](https://github.com/electron/electron/issues/1344#issuecomment-171516636) before I found this issue:

```
    var shell = require('electron').shell;
[...]

Issue #65, opened 2017-03-21
Opening link in a new tab


gavin-johnson84:
Is it possible to add `target="blank"` to Markdown links?

For example I wish to have the following URL open in a new browser tab when I left click:
`[link to website](http://www.website.com)`

Cheers

rexxars:
You can use a custom renderer:

```js
function LinkRenderer(props) {
  return <a href={props.href} target="_blank">{props.children}</a>
}

<ReactMarkdown
  source="Click [here](https://espen.codes/) to visit external site"
  renderers={{Link: LinkRenderer}}
/>
```

Dictory:
@rexxars `renderers={{Link: LinkRenderer}}` is incorrect, `Link:` => `link:` to lowercase =)

abourget:
it works here with the capital `L` in `Links:`

chenop:
I'm re-opening to suggest a syntax that opens a link in a new tab:
[link](url){:target="_blank"} 

Is this feature make sense? 
I'm willing to give it a try implementing this.

joelbowen:
For anyone else that ends up here via google search (like I did) and didn't notice this in the docs, you can now use the `props` to achieve this like so:

```jsx
<ReactMarkdown linkTarget="_blank" children={...} />
```

See [the docs](https://github.com/remarkjs/react-markdown#props) for more info.

phongapp:
> For anyone else that ends up here via google search (like I did) and didn't notice this in the docs, you can now use the `prop` `linkTarget` to achieve this like so:
> 
> ```js
> <ReactMarkdown linkTarget="_blank" children={...} />
> ```
> 
> See [the docs](https://github.com/remarkjs/react-markdown#props) for more info.

Does React Markdown automatically add `rel=”noopener noreferrer”` when using `linkTarget="_blank"` for secure?

ChristianMurphy:
This answer is pretty outdated, and should no longer be used as a reference.
The current recommendation is to use https://github.com/rehypejs/rehype-external-links which can apply `rel`.

I'm locking this thread as outdated.
Feel free to take follow up questions to the discussion forum https://github.com/orgs/remarkjs/discussions

PR #761, opened 2023-08-24
Added default value to link's "rel" when link's "target" is "_blank"

galabra:
### Initial checklist

### Description of changes

When a link's `target` is set to `_blank`, the consensus is adding `rel="noopener noreferrer"`.\
I added this as the default behavior and also allowed overriding this value.

This was discussed in the past: https://github.com/remarkjs/react-markdown/issues/413 and in the prehistoric https://github.com/remarkjs/react-markdown/issues/12

### DIFF
--- a/lib/ast-to-react.js
+++ b/lib/ast-to-react.js
@@ -97,6 +97,7 @@
  * @property {null|false|TransformLink} [transformLinkUri]
  * @property {TransformImage} [transformImageUri]
  * @property {TransformLinkTargetType|TransformLinkTarget} [linkTarget]
+ * @property {string} [linkRel]
  * @property {Components} [components]
  */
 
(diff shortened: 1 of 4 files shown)


wooorm:
Did you see the related issues when searching for `rel`?
https://github.com/remarkjs/react-markdown/pull/350 or more recently for example https://github.com/remarkjs/react-markdown/issues/65#issuecomment-1685086114.

What is your aversion to using a plugin instead of passing options?

galabra:
@wooorm Woah, I wasn't looking thoroughly enough and didn't see [#350](https://github.com/remarkjs/react-markdown/pull/350).
For simplicity sake, I honestly assumed the consensus and should be the default option here.

wooorm:
It’s an old issue from before plugins worked. They are closed because there is a better alternative.
The defaults are safe: don‘t use `_target`.
If anything, I’d remove `linkTarget`. It’s a bad idea anyway; I think it’s better to let users choose where they open things. E.g., https://css-tricks.com/use-target_blank/.

galabra:
I beg to differ 🙂 
Installing additional dependencies is never a good thing, but a tradeoff.
What good reason is there to **not** extend this functionality?

wooorm:
The trade off is between a small project that everyone uses or a big project with features many people do not use.
Another good reason is that people can already do this: pass an `a` component.
Another reason is that users will want to do different things.

wooorm:
Glad you found something that works, and thanks for the PR!

PR #796, opened 2023-11-12
fix: treat blob: urls as safe url

erfanium:
### Initial checklist

### Description of changes

I use react-markdown to preview an text input when images have not been uploaded to server. so it would be nice to be able to render blob images


wooorm:
You are proposing to change the defaults for everyone. Blobs are not always safe. You can already pass different things if you want that. Pass `urlTransform` in options

PR #799, opened 2023-11-15
Remove support for `className` prop

### LINKED ISSUE #781: Remove support for `className` prop
### Initial checklist

- [X] I read the [support docs](https://github.com/remarkjs/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/remarkjs/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/remarkjs/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Aremarkjs&type=Issues) and couldn’t find anything (or linked relevant results below)

### Problem

Typically a `className` prop in React adds a class name to the element it creates. For `react-markdown` this is not the case. Instead, it creates a wrapper that gets the class name. This is a fairly useless feature, because the user can wrap the content in a wrapper component themselves. That even gives the user more control.

In other words, these two are equivalent:

```jsx
import Markdown from 'react-markdown'

export function Component({ children }) {
  return (
    <Markdown className="some-class">
      {children}
    </Markdown>
  )
}
```

```jsx
import Markdown from 'react-markdown'

export function Component({ children }) {
  return (
    <div className="some-class">
      <Markdown>
        {children}
      </Markdown>
    </div>
  )
}
```

### Solution

Deprecate the `className` prop, and remove it in then next semver major release.

### Alternatives

- Do nothing.
- Don’t deprecate it.
- Make a semver major release just for this.

remcohaszing:
### Initial checklist

### Description of changes

This deprecates the `className` prop, but does not remove it yet. Users should wrap the `<Markdown>` component inside a `<div>` manually instead.

Closes #781

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,23 +2,6 @@
 
 All notable changes will be documented in this file.
 
-## 9.0.1
-
-### Deprecate `className`
-
-The `className` prop was deprecated.
-Not only does the `className` prop add the `className` prop, it also implies a
-wrapper `<div>` will be created.
-Instead, you should wrap the content in a `<div>` manually.
-
-```jsx
-<div className={className}>
-  <Markdown>
-    {children}
-  </Markdown>
-</div>
-```
-
 ## 9.0.0 - 2023-09-27
 
 *   [`b67d714`](https://github.com/remarkjs/react-markdown/commit/b67d714)
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,23 @@
 
 All notable changes will be documented in this file.
 
+## 9.0.1
+
+### Deprecate `className`
+
+The `className` prop was deprecated.
+Not only does the `className` prop add the `className` prop, it also implies a
+wrapper `<div>` will be created.
+Instead, you should wrap the content in a `<div>` manually.
+
+```jsx
+<div className={className}>
+  <Markdown>
+    {children}
+  </Markdown>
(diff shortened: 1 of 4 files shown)


JounQin (changes requested):
https://github.com/remarkjs/react-markdown/issues/781#issuecomment-1813496135

remcohaszing:
I misinterpreted the meaning of the word _deprecated_ here originally in a late evening.

I interpret _deprecated_ as _planned for change or removal_ with a warning. In this project _deprecated_ means _removed_ with an error, making deprecations semver major.

remcohaszing:
Sure :+1: 

It just feeld a bit weird to link a non-existent header from the code.

wooorm:
Ah that’s why, hmm, then I’d do a `<!-- To do: add heading for slug `deprecate-classname` -->`
Or these two comments might be enough!

remcohaszing:
Since this change is breaking, this PR shouldn’t be merged until the next major anyway. I think this PR thread is plenty reminder once this gets merged. :)

wooorm (approved):

ChristianMurphy (approved):

wooorm:
Released in [`10.0.0`](https://github.com/remarkjs/react-markdown/releases/tag/10.0.0)!

PR #836, opened 2024-06-28
Refactor to use `@import` to import types

remcohaszing:
### Initial checklist

### Description of changes

This replaces `@typedef` tags with `@import` where applicable. `@import` tags were introduced in TypeScript 5.5.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,15 +2,11 @@
 /// <reference types="mdast-util-to-hast" />
 
 /**
- * @typedef {import('hast').Element} Element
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -2,15 +2,11 @@
 /// <reference types="mdast-util-to-hast" />
```
wooorm, line 2: This one too?

remcohaszing: Just removing it turned out to be fine.

wooorm (approved):

PR #837, opened 2024-06-28
Replace `JSX` global with explicit `ReactElement`

remcohaszing:
### Initial checklist

### Description of changes

This makes the `react-markdown` types compatible with React 19. Dependencies still rely on the `JSX` namespace, but let’s trackle one thing at a time.

The better return type would be `ReactNode`, not `ReactElement`. But that breaks compatibility with TypeScript < 5.1. Let’s do that in a semver major.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,6 +1,7 @@
 /**
  * @import {Element, ElementContent, Nodes, Parents, Root} from 'hast'
  * @import {Components as JsxRuntimeComponents} from 'hast-util-to-jsx-runtime'
+ * @import {ReactElement} from 'react'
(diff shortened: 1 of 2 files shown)


wooorm (commented):
habibi please use separate prs for separate things, there’s an important change here, with a big diff of unrelated changes?

remcohaszing:
The second commit was needed, because apparently `type-coverage` started failing.

wooorm:
It was likely possible to side-step that with less changes: It didn’t get `ReturnType<typeof Markdown>`, for some reason, presumably `JSX.Element`, `ReactNode`, `ReactElement`, would’ve worked. Or a comment to ignore that one line?

PR #838, opened 2024-07-02
Leverage TypeScript project references

remcohaszing:
### Initial checklist

### Description of changes

This changes the TypeScript configuration of the project. It’s based on https://github.com/orgs/unifiedjs/discussions/238. The project now uses 2 TypeScript configurations: `tsconfig.json` and `tsconfig.build.json`. There’s also `tsconfig.base.json`. This is an implementation detail, it only exists so it can be extended from.

`tsconfig.build.json` is used to build the source files.

- It only includes the `lib` directory.
- It emits the type declarations into the `types` directory.
- It does not include the `node` types, meaning that importing Node.js builtins will cause the build to fail.
- It includes the `dom` lib, which is needed because of an upstream dependency.
- It does not allow the use of JSX.
- It includes a new file `lib/exports.ts`, which can be used to re-export additional types. This replaces the old `index.js` in the project root, which `@typedef` tags to mimic type exports.

`tsconfig.json` is used to typecheck the rest of the project.

- It excludes the `lib` directory.
- It does **not** emit type declarations.
- It includes the `node` types, meaning we can import Node.js builtins (such as `node:test`)
- It does **not** include the `dom` types, meaning we can’t use browser globals there.
- It builds `tsconfig.build.json` first, based on `references`.
- It type checks the `types` directory emitted by `tsconfig.build.json`. We have run into JSDoc specific emit issues before, so this is really nice to have.

To build or rebuild the project, we can now run either `tsc --build` or `tsc --build --force`. This correctly uses incremental builds, so a second run of `tsc --build` is faster.

Without this PR, after building, we have type errors in our editor. This is now solved.

Errors in tsconfig.json

Variations of this are possible. The main point is:

- Separate JavaScript source files from other JavaScript files.
- Use separate configurations for different environments/purposes. This is even more apparent for mono repos that target different conflicting environments. (I have a commit ready for https://github.com/mdx-js/mdx, but wanted to discuss this in a simpler repo first.)
- Use `rootDir` and `outDir`.
- Avoid writing `.d.ts` files.
- Use `.ts` files to write TypeScript things that are not possible with types in JSDoc.

An interesting idea that builds on this, is to move type definitions from `@typedef` tags into TypeScript files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,11 +69,13 @@
   ],
   "sideEffects": false,
   "type": "module",
-  "exports": "./index.js",
+  "exports": {
+    "types": "./types/exports.d.ts",
+    "default": "./lib/index.js"
+  },
   "files": [
     "lib/",
-    "index.d.ts",
-    "index.js"
+    "types/"
   ],
   "dependencies": {
     "@types/hast": "^3.0.0",
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,8 @@
 coverage/
 node_modules/
 types/
-*.tsbuildinfo
 *.log
+*.tsbuildinfo
 .DS_Store
 react-markdown.min.js
 yarn.lock
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 coverage/
 node_modules/
-*.d.ts
+types/
+*.tsbuildinfo
 *.log
 .DS_Store
 react-markdown.min.js
--- a/index.js
+++ b/index.js
@@ -1,9 +1,0 @@
-/**
- * @typedef {import('hast-util-to-jsx-runtime').ExtraProps} ExtraProps
- * @typedef {import('./lib/index.js').AllowElement} AllowElement
- * @typedef {import('./lib/index.js').Components} Components
- * @typedef {import('./lib/index.js').Options} Options
- * @typedef {import('./lib/index.js').UrlTransform} UrlTransform
- */
-
-export {Markdown as default, defaultUrlTransform} from './lib/index.js'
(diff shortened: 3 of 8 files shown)


package.json
```diff
@@ -69,11 +69,13 @@
   ],
   "sideEffects": false,
   "type": "module",
-  "exports": "./index.js",
+  "exports": {
+    "types": "./types/exports.d.ts",
+    "default": "./lib/index.js"
+  },
```
wooorm, line 75: Is it correct that the `types` field here is only needed because you want to overwrite `lib/index.d.ts`, for consumers of `'react-markdown'`?

package.json
```diff
@@ -69,11 +69,13 @@
...
   "files": [
     "lib/",
-    "index.d.ts",
-    "index.js"
+    "types/"
```
wooorm, line 78: What does the generated folder structure look like?
I am not happy about the extra folder. Feels like this is patching a bug that TS should solve.

wooorm:
OK, it’s ages so I forgot everything, but I checked the branch out, some things I don’t like:

- ts files
- special types folder for generates files
- 3 tsconfigs

While it is nice that typescript now has less problems in VS Code, it feels like a lot of bandaid to work around typescript not working well. I think it is reasonable to expect typescript to be able to generate `d.ts` and `.map` files next to `.js` files and for it to be possible to use typescript without `.ts` files

wooorm:
I appreciate thinking about how to improve things.
But my goal is to write JavaScript that runs. No compile step. JavaScript is a cool language: it has somewhat open governance.
I do not want TypeScript. TS has useful features: types. But is not a good language: it is owned by a giant corporation with a [no-good track record in open source](https://en.wikipedia.org/wiki/Embrace,_extend,_and_extinguish).

Issue #854, opened 2024-08-25
Dynamic children break the component


jonathandewitt-dev:
### Initial checklist

- [X] I read the [support docs](https://github.com/remarkjs/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/remarkjs/.github/blob/main/contributing.md)
- [x] I agree to follow the [code of conduct](https://github.com/remarkjs/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Aremarkjs&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

react-markdown

### Steps to reproduce

Occasionally, through some whimsical magic, React will choose to render a single string passed to `children` as an array containing one string. I can't seem to reproduce this scenario reliably, it just happens at what seems like random times.

This being the case, however, means that there are valid markdown strings being passed as children which this component is not resilient enough to handle.  I keep getting variations of the error message:

```
Unexpected value `test` for `children` prop, expected `string`
```

Where "test" was simply my text string passed as the sole child to the Markdown component. Upon inspection, I was able to discover that my `string` was being converted to `string[]` and then rejected by the package. Not a great experience because there's quite literally nothing that can be done within my control when facing this situation.

I was able to fix the problem by temporarily modifying this library's code in my node_modules, from this:

```ts
  if (typeof children === 'string') {
    file.value = children
  } else {
    unreachable(
      'Unexpected value `' +
        children +
        '` for `children` prop, expected `string`'
    )
  }
  ```
  
  to this:

```ts
  const allChildrenAreStrings = Array.isArray(children) ? children.every((child) => typeof child === 'string') : typeof children === 'string';

  if (allChildrenAreStrings) {
    file.value = Array.isArray(children) ? children.join('\n') : children
  } else {
    unreachable(
      'Unexpected value `' +
        children +
        '` for `children` prop, expected `string`'
    )
  }
```

### Expected behavior

It should handle `string[]` children by confirming that each child is indeed of type `string` and then `Array.join('\n')` them to continue processing the markdown.

### Actual behavior

For `string[]` arrays, it reports an error and crashes the app.

wooorm:
Hi!

`children` is the value *you* pass in. Sometimes through JSX. But it’s your choice how to pass it. You can pass weird things: `<Markdown>{1}{2}{3}</Markdown>`.

[...]

PR #855, opened 2024-08-25
Fixing `string[]` scenario

jonathandewitt-dev:
### Initial checklist

### Description of changes

This addresses issue #854 by adding support for `string[]` in addition to `string` children.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -162,8 +162,10 @@ export function Markdown(options) {
 
   const file = new VFile()
 
-  if (typeof children === 'string') {
-    file.value = children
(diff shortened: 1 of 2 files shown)


remcohaszing:
I think this only adds ambiguity. We can better make it clear that we only accept a string. If you have an array of strings you want to join together using newlines, you can write that in your own code.

wooorm:
Thanks but lets first discuss the issue more: https://github.com/remarkjs/react-markdown/issues/854.

ChristianMurphy:
resolved at https://github.com/remarkjs/react-markdown/issues/854#issuecomment-2337579770

PR #870, opened 2024-10-28
Fix Object.hasOwn is not a function

andylin9:
### Initial checklist

* [x] I read the support docs 
* [x] I searched issues and couldn’t find anything (or linked relevant results below) 
* [x] If applicable, I’ve added docs and tests

### Description of changes

To support old version of browser, add the polyfill of Object.hasOwn.
About Object.hasOwn browser compatibility: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn#browser_compatibility

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -162,6 +162,11 @@ export function Markdown(options) {
 
   const file = new VFile()
 
+  if (!Object.hasOwn) {
+  Object.hasOwn = (obj, prop) =>
+    Object.prototype.hasOwnProperty.call(obj, prop);
+  }
+
   if (typeof children === 'string') {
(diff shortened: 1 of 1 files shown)


remcohaszing:
This package [targets Node.js 16 or greater](https://github.com/remarkjs/react-markdown?tab=readme-ov-file#compatibility). This means [`Object.hasOwn` is available](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn#browser_compatibility). If you want to support older environments, you can provide the polyfill in your own project.

PR #878, opened 2024-12-20
Separate all typedefs into their own JSDoc blocks

remcohaszing:
### Initial checklist

* [x] I read the support docs 

### Description of changes

This change separates all typedefs into their own JSDoc blocks. Having them in the same block, can be problematic sometimes. For example when they contain `@template` tags.


ChristianMurphy (approved):
I don't have strong feelings either way, I'm fine with this change, though I think things currently work without it too.

PR #879, opened 2024-12-20
Add support for React 19

### LINKED ISSUE #877: React 19 support
Followup to https://github.com/remarkjs/react-markdown/issues/871

`react-markdown` relies on `JSX` being in the global namespace, which has been removed in `@types/react^19`. 

Per @ChristianMurphy [here](https://github.com/remarkjs/react-markdown/issues/871#issuecomment-2551504752):

> A new issue would be better, we won't support React 19 RC, only GA.

[React 19 is now GA](https://react.dev/blog/2024/12/05/react-19), so this issue is to track.

remcohaszing:
### Initial checklist

* [x] I read the support docs 

### Description of changes

This adds support for React 19 types, by removing the dependency on the types provided by hast-util-to-jsx-runtime, thus removing the dependency on the JSX global namespace.

Unfortunately the current setup doesn’t really allow testing the types against multiple React versions. #838 would enable that.

Closes #877

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,7 +1,6 @@
 /**
  * @import {Element, ElementContent, Nodes, Parents, Root} from 'hast'
- * @import {Components as JsxRuntimeComponents} from 'hast-util-to-jsx-runtime'
- * @import {ReactElement} from 'react'
+ * @import {ComponentProps, ElementType, ReactElement} from 'react'
  * @import {Options as RemarkRehypeOptions} from 'remark-rehype'
  * @import {BuildVisitor} from 'unist-util-visit'
  * @import {PluggableList} from 'unified'
@@ -21,7 +20,16 @@
  */
 
 /**
- * @typedef {Partial<JsxRuntimeComponents>} Components
+ * @typedef ExtraProps
+ *   Extra fields we pass.
+ * @property {Element | undefined} [node]
+ *   passed when `passNode` is on.
+ */
+
+/**
+ * @typedef {{
+ *   [Key in Extract<ElementType, string>]?: ElementType<ComponentProps<Key> & ExtraProps>
+ * }} Components
  *   Map tag names to components.
  */
 
@@ -225,6 +233,9 @@ export function Markdown(options) {
 
   return toJsxRuntime(hastTree, {
     Fragment,
+    // @ts-expect-error
+    // React components are allowed to return numbers,
+    // but not according to the types in hast-util-to-jsx-runtime
     components,
     ignoreInvalidStyle: true,
     jsx,
--- a/index.js
+++ b/index.js
@@ -1,7 +1,7 @@
 /**
- * @typedef {import('hast-util-to-jsx-runtime').ExtraProps} ExtraProps
  * @typedef {import('./lib/index.js').AllowElement} AllowElement
  * @typedef {import('./lib/index.js').Components} Components
+ * @typedef {import('./lib/index.js').ExtraProps} ExtraProps
  * @typedef {import('./lib/index.js').Options} Options
  * @typedef {import('./lib/index.js').UrlTransform} UrlTransform
  */


lib/index.js
```diff
@@ -217,6 +233,9 @@ export function Markdown(options) {
 
   return toJsxRuntime(hastTree, {
     Fragment,
+    // @ts-expect-error
+    // React components are allowed to return numbers,
+    // but not according to the types in hast-util-to-jsx-runtime
```
ChristianMurphy, line 238: Is this related to https://github.com/syntax-tree/hast-util-to-jsx-runtime/pull/6 ?

ChristianMurphy (approved):
LGTM, thanks @remcohaszing

lib/index.js
remcohaszing: Somewhat yes. But also apparently `hast-util-to-jsx-runtime` doesn’t support `JSX.ElementType`, which JSX runtimes may optionally define. That adds another complexity on top of that issue.

sanskar-19:
@remcohaszing, What is the expected timeline for this release? - 

We are using `react-markdown` with Next14, and want to upgrade to Next15, 
which supports React 19, which is not currently supported by `react-markdown`.

bluepeter:
Let's make this happen!

wooorm:
Why does `Copilot` review things?

lib/index.js
```diff
@@ -18,10 +17,23 @@
...
+ *   Extra fields we pass.
+ * @property {Element | undefined} [node]
+ *   passed when `passNode` is on.
+ */
+
+/**
+ * @typedef {{
+ *   [Key in Extract<ElementType, string>]?: ElementType<ComponentProps<Key> & ExtraProps>
+ * }} Components
  *   Map tag names to components.
```
wooorm, line 33: I believe this type is not equivalent to the previous `Components`.
But this is listed as a patch.
Is it intentional that you remove `keyof JSX.IntrinsicElements`?

lib/index.js
```diff
@@ -62,7 +76,9 @@
  *   with `unwrapDisallowed` the element itself is replaced by its children.
  * @property {UrlTransform | null | undefined} [urlTransform]
  *   Change URLs (default: `defaultUrlTransform`)
- *
+ */
+
+/**
```
wooorm, line 81: I agree with these changes, but it would be good to have different commits for things of varying importance. This commit/PR is very important, many humans will look at the change, and see this mix of important/trivial.
Not blocking, but like: feel free to commit such refactors things directly to main (`Refactor code-style`), and use a PR for things that need eyes on them!

lib/index.js
wooorm: Would be nice to solve such things instead of `ts-expect-error`

ChristianMurphy:
> Why does Copilot review things?

What a very somber and philosophical question 😆
Why do you review things?
Why does anyone do anything?

---

If you are asking why it appeared on this PR as a reviewer.
It is because I invited it as an additional reviewer to see how it would do.

lib/index.js
remcohaszing: I agree. It’s why I created #878 first and based this change on that one. I’ll merge it.

lib/index.js
remcohaszing: `ElementType` includes `JSX.IntrinsicElements`.

lib/index.js
remcohaszing: I agree, but that issue has been proven hard to resolve upstream. (If it can be resolved upstream at all.) This `@ts-expect-error` is internal anyway, and this change unblocks a lot of users.

lib/index.js
wooorm: @remcohaszing No, I mean `keyof`? Does this allow `b: 'i'`

wooorm:
> If you are asking why it appeared on this PR as a reviewer.
> It is because I invited it as an additional reviewer to see how it would do.

And, are you happy with it? 😅

I am emboldened in my view that I am not happy with it 😅

lib/index.js
remcohaszing: It does! And there’s also a pre-existing test which covers this.

wooorm: Weird, I do not see it in the source

remcohaszing:
I’m fine with @ChristianMurphy wanting to try Copilot reviews, but IMO the result is really pointless, at least in this case.

lib/index.js
remcohaszing: https://github.com/remarkjs/react-markdown/blob/27d3949b31beb7aa7a6c0d3d4d34e6fd0965a7d3/test.jsx#L523

wooorm: right no that, I meant in the `@types/react` code

wooorm (approved):

lib/index.js
wooorm: It feels like it could be a bug that `@types/react` supports this today — that they could remove it somewhere in the future

wooorm: not sure where it happens but when I change that line to `h1: 'x-y'` we get a type error — which indicates that I should worry less because it is type checked — even though the types should likely allow custom elements like that.

ChristianMurphy:
> And, are you happy with it? 😅

It's not perfect, but getting better.

> I am emboldened in my view that I am not happy with it 😅

Okay

> the result is really pointless, at least in this case.

It was in this case.
It is getting less noisy and nitpicky which is good.
I wish there were a way to specify custom criteria.

cyberluke:
Why not reuse this solution: https://github.com/quantizor/markdown-to-jsx/commit/918b44b5b8e4e96afc7e4a36d73c9c8a926782e6#diff-4b76aeed93d86ab1e483e12333e1bae64ccdc1a1e4fd27bec2b32f41db08e6cfR5

and just change:
JSX.IntrinsicAttributes

to
React.JSX.IntrinsicAttributes

in complex-types.ts and complex-types.d.ts?

remcohaszing:
The goal of `hast-util-to-jsx-runtime` is to support any JSX runtime, not just React.

wooorm:
Released in [`9.0.2`](https://github.com/remarkjs/react-markdown/releases/tag/9.0.2)!

wooorm:
done in [`9.0.3`](https://github.com/remarkjs/react-markdown/releases/tag/9.0.3)!

ChristianMurphy:
@ludobonnet .d.ts are aren't checked into source, so they won't show in the diff, they are artifacts on npm.

PR #891, opened 2025-02-17
Avoid bad hooks dependency usage in MarkdownHooks

remcohaszing:
### Initial checklist

* [x] I read the support docs 

### Description of changes

As some comments point out in #890, there are some bad React hook usages in those changes. A `useEffect()`, `useMemo()`, or `useCallback()` should always specify all of its dependencies in the dependency array. But the dependencies need to be referentially equal in order to not trigger the hook, which is often not possible, given that plugin arrays are often created during render.

This change replaces the `useEffect()` logic with a `useRef()`. Using `useRef()`, we can persist an object instance across rerenders. We can then perform our own props comparison logic to update or not instead of relying on a dependency array. We also avoid `setState()`, which triggered another render and could cause flashes of empty content.

The processor is updated if `remarkRehypeOptions`, `rehypePlugins`, or `remarkPlugins` changed. The promise is updated if the processor or `children` has changed.

#### TODO

- [ ] Deep comparison of plugin options? Or to some depth?
- [ ] Deep comparison of `remarkRehypeOptions`??
- [ ] Add tests?


ChristianMurphy (approved):

remcohaszing:
I realized the `use` API was introduced in React 19. We could choose to update the peer dep, which would be a breaking change. Or we could require React 19 for `MarkdownHooks` only. Or we may need to look for an alternative to resolve the promise after all.

wooorm (commented):
I think it is sad that we would implement such caching equivalence mechanisms; I believe that is what frameworks are supposed to do for you.
As you know, caching is difficult. That shows here too as there are several problems with this equivalence function: a) plugins support several parameters not just options; b) presets.
But, most importantly: you only have a `return true` on L464: you can remove the entire function as it will always be the tail `return false`, just with more or less work.

remcohaszing:
> I think it is sad that we would implement such caching equivalence mechanisms; I believe that is what frameworks are supposed to do for you. As you know, caching is difficult.

I somewhat agree here. *Just React* works great, but it can be hard to integrate external libraries that weren’t built with the React mindset. This would be a lot simpler if we didn’t allow `MarkdownHooks` to accept arguments to configure a processor, but instead it would work on a pre-existing processor.

> That shows here too as there are several problems with this equivalence function: a) plugins support several parameters not just options; b) presets. But, most importantly: you only have a `return true` on L464: you can remove the entire function as it will always be the tail `return false`, just with more or less work.

Good finds! This is really going to need tests.

PR #898, opened 2025-03-06
Replace custom loader with @node-loaders/tsx

remcohaszing:
### Initial checklist

* [x] I read the support docs 

### Description of changes

We need a loader to load our JSX test file. We currently use a local script, but there are several packages that do this. We don’t need to maintain our own.

This replaces our custom loader with [`@nodejs-loaders/tsx`](https://github.com/nodejs-loaders/nodejs-loaders/tree/main/packages/tsx) by the Node.js team.


remcohaszing:
Apparently this doesn’t work with all Node.js versions we support.

ChristianMurphy (approved):
Looks good to me, let's merge this when hydrogen is supported, or when we drop hydrogen support

JounQin:
What's the different with using `tsx` itself?

Issue #865, opened 2024-10-17
tel: href tag not working


vitalijalbu:
### Initial checklist

- [X] I read the [support docs](https://github.com/remarkjs/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/remarkjs/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/remarkjs/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Aremarkjs&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

9.0.1

### Steps to reproduce

Basically I added a simple phone tel: link and it not passing href props

![Screenshot 2024-10-17 alle 10 40 17](https://github.com/user-attachments/assets/c5b8a47b-363d-40f5-a2f1-25ecfe301295)

### Expected behavior

It should work as expected

### Actual behavior

Not passing props.

my code:

  ```
const components = {
    p: PComponent,
    h1: H1Component,
    h2: H2Component,
    h3: H3Component,
    h4: H4Component,
    h5: H5Component,
    h6: H6Component,
    strong: StrongComponent,
    em: EmComponent,
    a: AComponent,
    ul: UlComponent,
    ol: OlComponent,
    img: ImgComponent,
    blockquote: QuoteComponent,
    table: TableComponent,
  };
  
  <Markdown
        remarkPlugins={[
          remarkGfm,
          remarkCustomHeaderId,
          remarkParse,
          remarkLinkTarget,
        ]}
        rehypePlugins={[rehypeRaw]}
        components={components}
      >
        {markdown}
      </Markdown>
  
function AComponent({ children, href = '#', ...props }: LinkProps) {
  console.log('AComponent', {href, props, children});
  if (href.startsWith('tel:')) {
    return (
      <a href={href} {...props}>
        {children}
      </a>
    );
  } else {
    return (
      <TextLink href={href} {...props}>
        {children}
      </TextLink>
    );
  }
}

```

### Affected runtime and version

node 20

vitalijalbu:
Tried also on your md editr online and not working

wooorm:
It is working. It is removed. Read the docs. You want to define `urlTransform`.

Issue #829, opened 2024-05-17
React-markdown isn't rendering phone number links anymore


aligreene:
### Initial checklist

- [X] I read the [support docs](https://github.com/remarkjs/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/remarkjs/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/remarkjs/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Aremarkjs&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

react-markdown v9.0.0

### Link to runnable example

https://codesandbox.io/p/devbox/react-markdown-telephone-link-bug-3vsqry

### Steps to reproduce

We recently updated from react-markdown ^8.0.3 to ^9.0.0 and since then, any phone link renders with no value in the link href.

See sandbox for runnable example, or follow steps below:

1. Create a markdown telephone link e.g.:
`[123 456 789](tel:123456789)`
2. Check the output.

### Expected behavior

The link is rendered in HTML with a href value:
`<a href="tel:123456789">123 456 789</a>`

### Actual behavior

The link renders without a href value:
`<a href="">123 456 789</a>`

wooorm:
See the changelog: https://github.com/remarkjs/react-markdown/blob/main/changelog.md#add-urltransform.

aligreene:
Thanks very much

Issue #655, opened 2021-11-10
Support Autolinking Phone Numbers


steverecio:
### Initial checklist

- [X] I read the [support docs](https://github.com/remarkjs/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/remarkjs/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/remarkjs/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Aremarkjs&type=Issues) and couldn’t find anything (or linked relevant results below)

### Problem

There is support to autolinking URLs but phone numbers have to be manually linked, ie: `[+1 (800) 111-1111](tel:1800111111)`

### Solution

Would be great if phone numbers are autolinked similarly to URLs

### Alternatives

The alternatives are building the links directly with markdown syntax or creating another plugin for this functionality

wooorm:
You can create a plugin.
It’s not standard markdown.
This project is about  standard markdown.
You can do many other things with plugins.

ChristianMurphy:
To add on, the unified (remark) website includes guides introducing plugins https://unifiedjs.com/learn/
And there are some plugins which could help inspire how it could work like remark-gemoji https://github.com/remarkjs/remark-gemoji

PR #904, opened 2025-04-07
Add tel: as href support

ValeryVerkhoturov:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Add support tel as href:

```jsx
<Markdown children="[](tel:+12345678910)" />
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -121,7 +121,7 @@ const changelog =
 const emptyPlugins = []
 /** @type {Readonly<RemarkRehypeOptions>} */
 const emptyRemarkRehypeOptions = {allowDangerousHtml: true}
-const safeProtocol = /^(https?|ircs?|mailto|xmpp|tel)$/i
+const safeProtocol = /^(https?|ircs?|mailto|tel|xmpp)$/i
 
 // Mutable because we `delete` any time it’s used and a message is sent.
 /** @type {ReadonlyArray<Readonly<Deprecation>>} */
--- a/lib/index.js
+++ b/lib/index.js
@@ -121,7 +121,7 @@ const changelog =
 const emptyPlugins = []
 /** @type {Readonly<RemarkRehypeOptions>} */
 const emptyRemarkRehypeOptions = {allowDangerousHtml: true}
-const safeProtocol = /^(https?|ircs?|mailto|xmpp)$/i
+const safeProtocol = /^(https?|ircs?|mailto|xmpp|tel)$/i
 
 // Mutable because we `delete` any time it’s used and a message is sent.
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -121,7 +121,7 @@ const changelog =
 const emptyPlugins = []
 /** @type {Readonly<RemarkRehypeOptions>} */
 const emptyRemarkRehypeOptions = {allowDangerousHtml: true}
-const safeProtocol = /^(https?|ircs?|mailto|xmpp)$/i
+const safeProtocol = /^(https?|ircs?|mailto|xmpp|tel)$/i
```
JounQin, line 124:
```suggestion
const safeProtocol = /^(https?|ircs?|mailto|tel|xmpp)$/i
```

Let's sort the protocols in the `RegExp`.

JounQin (changes requested):

ChristianMurphy:
Tel is intentionally not supported https://github.com/remarkjs/react-markdown/issues/865, https://github.com/remarkjs/react-markdown/issues/829, and https://github.com/remarkjs/react-markdown/issues/655#issuecomment-965649809, but can be added through an option

JounQin:
https://spec.commonmark.org/0.22/#scheme

@ChristianMurphy It seems there are a lot of protocols supported in commonmark spec.

ChristianMurphy:
There are a lot that are _parsed_, but they are intentionally filtered out to make renders safe, for example GFM e.g. `<tel:+12345678910>` renders as <tel:+12345678910>

JounQin:
@ChristianMurphy Thanks for clarification, just found the related commit https://github.com/remarkjs/react-markdown/commit/08ead9ef38765d7b273065edb8d72ab2aa5ab512

But it lacks explanation why it's removed and considered as unsafe, and why `gfm` is considered.

wooorm:
The default is what GH does (not particularly GFM, that’s only *a part* of *everything* that GH does). GH has a small allow list. That small list is safe. We follow GH in many cases.

You can configure things: see the readme: pass a `urlTransform` option!

JounQin:
> The default is what GH does (not particularly GFM, that’s only _a part_ of _everything_ that GH does). GH has a small allow list. That small list is safe. We follow GH in many cases.
> 
> You can configure things: see the readme: pass a `urlTransform` option!

Maybe it's worth to mention that we follow GitHub's behavior as default at https://github.com/remarkjs/react-markdown#security

wooorm:
Thanks for the suggestion. I added a paragraph on the internal workings of `defaultUrlTransform`. I think it fits better there!

PR #907, opened 2025-04-16
[DevEx Improvement] Remove default export

pastacolsugo:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Hello everyone,
thank you for your work on `react-markdown`!

I'm submitting this PR to propose a Developer Experience improvement, by removing the default export in `index.js`.

[The current code](https://github.com/remarkjs/react-markdown/blob/8545ebd7416beb2af2bfef59440db231abe69f13/index.js#L13) exports the `Markdown` component (synchronous version) as default.

This default export makes it possible to import the module by doing:

```typescript
import Markdown from 'react-markdown';
```

In my case I then realised I needed `MarkdownHooks`, thus changed my import as usual, to:

```typescript
import MarkdownHooks from 'react-markdown';
```

And here lies the catch: I'm still importing the `Markdown` component!

I forgot to add curly braces, and the default export happily worked to import the `Markdown` in the confusingly named `MarkdownHooks` variable.

Removing the default export would force the developer to import using curly braces from the beginning, avoiding this confusing mistake. It's also consistent with how most other packages work.

This PR removes the default export, and updates the docs accordingly.

[This article](https://phuoc.ng/collection/tips/avoid-using-default-exports/) explains other benefits of removing default exports better than I could.

Cheers! 🎲


remcohaszing:
I also prefer named exports. But this is definitely a breaking change. I don’t think we should release a semver major just for this.

wooorm:
hello! I would agree with you if there were few users.

I am very strongly against this. You are asking 1234567 people to change their code. For no real benefit. Also, there are 10 years of examples out there: LLMs will keep outputting the old version that doesn’t work for years. As you show in your use case, many users of this package are beginners and will mix things up.

backcompat is nice. We don’t have to break things. Sometimes the world isn’t perfect, that’s ok

PR #908, opened 2025-04-18
Create processor in useEffect in MarkdownHooks

remcohaszing:
### Initial checklist

* [x] I read the support docs 

### Description of changes

In `MarkdownHooks` we don’t need to create the processor for every render. We only need it inside the effect.


remcohaszing:
Not necessarily, but it does give me an idea. I need to fiddle a bit.

remcohaszing:
I’m closing this in favor of #909

PR #909, opened 2025-04-18
Memoize the processor in MarkdownHooks

remcohaszing:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Some plugins may perform heavy work in preparation to make the transformer light-weight. A good example of this is `rehype-starry-night`.

Without this change, an interactive markdown editor which includes `rehype-starry-night` is very sluggish. With this change, performance is as good as if `rehype-starry-night` isn’t even used.

Fix https://github.com/orgs/rehypejs/discussions/194

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -213,7 +213,9 @@ export async function MarkdownAsync(options) {
  */
 export function MarkdownHooks(options) {
   const processor = useMemo(
-    () => createProcessor(options),
+    function () {
+      return createProcessor(options)
+    },
     [options.rehypePlugins, options.remarkPlugins, options.remarkRehypeOptions]
   )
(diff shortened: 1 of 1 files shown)


JounQin (approved):
LGTM

lib/index.js
```diff
@@ -212,7 +212,10 @@ export async function MarkdownAsync(options) {
  *   React node.
  */
 export function MarkdownHooks(options) {
-  const processor = createProcessor(options)
+  const processor = useMemo(
+    () => createProcessor(options),
```
wooorm, line 216: please use a regular function, thanks

wooorm (approved):

wooorm:
Thanks! :)

PR #913, opened 2025-05-06
Add onComplete callback for completing processing in MarkdownHooks

### LINKED ISSUE #912: onComplete in MarkdownHooks
### Initial checklist

- [x] I read the [support docs](https://github.com/remarkjs/.github/blob/main/support.md)
- [x] I read the [contributing guide](https://github.com/remarkjs/.github/blob/main/contributing.md)
- [x] I agree to follow the [code of conduct](https://github.com/remarkjs/.github/blob/main/code-of-conduct.md)
- [x] I searched [issues](https://github.com/search?q=user%3Aremarkjs&type=issues) and [discussions](https://github.com/orgs/remarkjs/discussions) and couldn’t find anything (or linked relevant results below)

### Problem

Hi, I'm the developer of [react-markdown-typewriter](https://github.com/DRincs-Productions/react-markdown-typewriter) a library that adds elements that give the possibility to animate each letter of the markdown component, by default, to create a typewriter animation. the animations are created with motion.

For now I have created the variants for the Markdown and MarkdownAsync components.

I'm trying to create the variant for MarkdownHooks too.

Roughly how it works is the following, a parent component (motion) starts the animation of each child (Markdown elements) every n milliseconds. the animation must be triggered by an event for example during the first render.

With MarkdownHooks if I trigger the animation during the first render it will be executed on the fallback element and not when the MarkdownHooks element has finished loading.

So I would need a function (onComplete) that warns me that the MarkdownHooks loading is finished and then triggers the animation.

### Current solutions

I've currently solved it this way, although it can't be considered a solution 😒

```
    useEffect(() => {
        setTimeout(() => {
            // start animation
        }, 10);
    }, [text]);
```

### Proposed solutions

add a function that is launched when markdown loading is finished

```
            <MarkdownHooks
                onComplete={() => {
                      // start animation
}}
            >
                {text}
            </MarkdownHooks>
```

BlackRam-oss:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Add onComplete callback for completing processing in MarkdownHooks

close #912

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -246,7 +246,7 @@ export function MarkdownHooks(options) {
         cancelled = true
       }
     },
-    [options.children, processor]
+    [options.children, processor, options.onComplete]
   )
 
   if (error) throw error
--- a/lib/index.js
+++ b/lib/index.js
@@ -82,6 +82,8 @@
  *   Configuration specifically for {@linkcode MarkdownHooks}.
  * @property {ReactNode | null | undefined} [fallback]
  *   Content to render while the processor processing the markdown (optional).
+ * @property {function(): void | null | undefined} [onComplete]
+ *   Callback when the processor is done processing.
  */
 
 /**
@@ -232,6 +234,7 @@ export function MarkdownHooks(options) {
         if (!cancelled) {
           setError(error)
           setTree(tree)
+          options.onComplete?.()
         }
       })
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -82,7 +82,7 @@
  *   Configuration specifically for {@linkcode MarkdownHooks}.
  * @property {ReactNode | null | undefined} [fallback]
  *   Content to render while the processor processing the markdown (optional).
- * @property {function(Root | undefined, Error | undefined): void | null | undefined} [onComplete]
+ * @property {(file: VFile, hast: Root | undefined, error: Error | undefined) => unknown} [onComplete]
  *   Callback when the processor is done processing.
  */
 
@@ -234,7 +234,7 @@ export function MarkdownHooks(options) {
         if (!cancelled) {
           setError(error)
           setTree(tree)
-          options.onComplete?.(tree, error)
+          options.onComplete?.(file, tree, error)
         }
       })
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -82,7 +82,7 @@
  *   Configuration specifically for {@linkcode MarkdownHooks}.
  * @property {ReactNode | null | undefined} [fallback]
  *   Content to render while the processor processing the markdown (optional).
- * @property {function(): void | null | undefined} [onComplete]
+ * @property {function(Root | undefined, Error | undefined): void | null | undefined} [onComplete]
  *   Callback when the processor is done processing.
  */
 
@@ -234,7 +234,7 @@ export function MarkdownHooks(options) {
         if (!cancelled) {
           setError(error)
           setTree(tree)
-          options.onComplete?.()
+          options.onComplete?.(tree, error)
         }
       })
 
--- a/test.jsx
+++ b/test.jsx
@@ -1167,6 +1167,30 @@ test('MarkdownHooks', async function (t) {
     assert.equal(result.container.innerHTML, '<p>a</p>')
   })
 
+  await t.test('should support `onComplete`', async function () {
+    const plugin = deferPlugin()
+    let succes = false
+    render(
+      <MarkdownHooks
+        children={'a'}
+        onComplete={function () {
+          succes = true
+        }}
+        rehypePlugins={[plugin.plugin]}
+      />
+    )
+
+    assert.equal(succes, false)
+
+    plugin.resolve()
+
+    await waitFor(function () {
+      assert.notEqual(succes, false)
+    })
+
+    assert.equal(succes, true)
+  })
+
   await t.test('should support plugins that error', async function () {
     const plugin = deferPlugin()
     const result = render(


lib/index.js
```diff
@@ -82,6 +82,8 @@
  *   Configuration specifically for {@linkcode MarkdownHooks}.
  * @property {ReactNode | null | undefined} [fallback]
  *   Content to render while the processor processing the markdown (optional).
+ * @property {function(Root | undefined, Error | undefined): void | null | undefined} [onComplete]
```
remcohaszing:
```suggestion
 * @property {(hast: Root | undefined, error: Error | undefined) => unknown} [onComplete]
```

Also, I think this should receive the VFile. Access to the file’s `messages` can be useful.

lib/index.js
```diff
@@ -232,6 +234,7 @@ export function MarkdownHooks(options) {
         if (!cancelled) {
           setError(error)
           setTree(tree)
+          options.onComplete?.(tree, error)
```
remcohaszing: This needs to be added to the `useEffect` dependency array.

remcohaszing (changes requested):
I’m not really for or against this change. I see why it’s useful, but I also wonder if we should add such features. `react-markdown` is a relatively simple, high-level library. All the pieces are available in the unified ecosystem to build a more specialized integration if the user needs it.

JounQin:
Please open a new issue before PR next time.

BlackRam-oss:
> Please open a new issue before PR next time.

Hello @JounQin 

I created an issue, but I'm afraid I didn't manage to link it https://github.com/remarkjs/react-markdown/issues/912

Anyway, yes, next time I start editing I'll wait for approval.

JounQin:
> I created an issue, but I'm afraid I didn't manage to link it #912
>
> Anyway, yes, next time I start editing I'll wait for approval.

Let's convert this PR into draft and discuss with other maintainers there then.

By the way, the CI is still broken.

BlackRam-oss:
@JounQin 

Unfortunately I was not able to understand what the CI error is.

All tests went well but it gives the following error:
`ERROR: Coverage for branches (99.55%) does not meet global threshold (100%)`

JounQin:
> All tests went well but it gives the following error:
>
> `ERROR: Coverage for branches (99.55%) does not meet global threshold (100%)`

You didn't provide a new test case for `options.onComplete` at all.

ChristianMurphy:
I won't block this, but am not sure I'd recommend it either.

I don't think JS is needed to make a typing animation, CSS can likely work well enough, see https://github.com/orgs/remarkjs/discussions/1321 and https://github.com/orgs/remarkjs/discussions/1310

> onComplete callback 

names like `onComplete` and `onDone` imply that they refer to a promise, but here it doesn't

> ERROR: Coverage for branches (99.55%) does not meet global threshold (100%)

The tests measure code coverage https://en.wikipedia.org/wiki/Code_coverage
Branch coverage, the measure of conditions in the code that are run by tests, is missing some cases from tests, so is failing.

wooorm:
If this was done, I: a) worry about the name; b) think that probably no parameters should be passed at all? c) there is no error handling; d) there should probably be a result handling?

But importantly, I worry what this is for. Have you actually tried to use this? It looks like it could be abused to change the DOM that we control. And I think this sets expectations to users that they can modify file/tree/error but it wouldn’t actually *do* what they wanted.

BlackRam-oss:
@ChristianMurphy 

> I don't think JS is needed to make a typing animation, CSS can likely work well enough, see https://github.com/orgs/remarkjs/discussions/1321 and https://github.com/orgs/remarkjs/discussions/1310

I switched from using CSS to using motion with https://motion.dev/docs/react-transitions#staggerchildren (which I believe still uses mostly CSS) because I needed to intercept some animation events (for example the completion of the animation of the entire text and certain elements of the text such as punctuation) and give the user the possibility to have total control of the animations in an easy way.
There is probably the possibility to achieve a similar result with CSS (with a lot of work), but I don't think I would have been able to give the possibility to the user of my library to be able to modify it in an easy way.

@wooorm 

> But importantly, I worry what this is for

In any case I imagine that also in other cases those who are using MarkdownHooks might need to know if the user has viewed the text or just the loading.

For example in a very common case that is the management of notifications (with a very long text), in particular if it has been read or not. Those who are using MarkdownHooks to display the text of messages might need to notify the bakend that the message has been read by the user only after MarkdownHooks has finished loading.

remcohaszing:
Note that in the callback, it hasn’t been rendered. Only the React state was just updated.

There are ways to actually know when it’s done rendering. For example, 

```jsx
import { Root } from 'hast'
import { createContext, ReactNode, Ref, use, useState } from 'react'

const RenderedContext = createContext<Ref<HTMLSpanElement>>()

function Span({ 'data-rendered': rendered, ...props }): ReactNode {
  return <span {...props} ref={rendered ? use(RenderedContext) : null} />
}

const components = {
  span: Span
}

function rehypeRenderedIndicator() {
  return (ast: Root) => {
    ast.children.push({
      type: 'element',
      tagName: 'span',
      properties: {
        'data-rendered': 'true'
      },
      children: [],
    })
  }
}

const rehypePlugins = [rehypeRenderedIndicator]

function App(): ReactNode {
  const [rendered, setRendered] = useState<HTMLSpanElement>()

  return (
    <RenderedContext value={setRendered}>
      <MarkdownHooks rehypePlugins={rehypePlugins}>{''}</MarkdownHooks>
    </RenderedContext>
  )
}
```

JounQin:
> There are ways to actually know when it’s done rendering. For example

Oh my eyes.

wooorm:
> For example in a very common case that is the management of notifications (with a very long text), in particular if it has been read or not. Those who are using MarkdownHooks to display the text of messages might need to notify the bakend that the message has been read by the user only after MarkdownHooks has finished loading.

This proposed feature is not a good solution for that use case

BlackRam-oss:
> This proposed feature is not a good solution for that use case

@wooorm, Why? Is the problem where the `onComplete` function is executed?

---

The fact of not giving the possibility to access the file, hast and error, and the fact that the name onComplete could be not apriopiated, I can understand and honestly for me it is indifferent.

However I did not understand if there are cons on the fact of adding a function to notify the developer of the completion.

cavalera76 (commented):
awesome

wooorm:
Sorry for the wait, thanks for your patience!

- `tree`, `file` are intended to remain internal. Use plugins or `unified` directly if you want to access them.
- this `onComplete` idea is for whether something has changed, it can be rendered many times, that is not related to whether a human has read something. You probably want to be explicit about that (a button or so), or when explicit, you’d need scroll position? but there are all these problems with that (what if users do not see?); you probably don‘t want to spam your backends for every character that someone writes into a `<Markdown>` component.
- there may be use cases to this feature, but the use case you have is not that, and I want to publish things that have good use cases, I don’t want to maintain things that don’t have real use cases, or put things in docs/tests that don‘t have good use cases!

PR #918, opened 2025-06-21
chore: add react native remark link

imwithye:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Add related link to react-native-remark. I believe many developers are looking for a good solution for using remark with React Native, but existing options are limited. I created a library to fill this gap, and I think it would be helpful to include it as a related link. Thank you!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -810,8 +810,8 @@ It lets you define your own schema of what is and isn’t allowed.
   — hook based alternative
 * [`rehype-react`][github-rehype-react]
   — turn HTML into React elements
-* [`react-native-remark`][github-react-native-remark]
-  - remark on React Native
+* [`react-native-remark`][github-react-native-remark] 
+  — remark on React Native
 
 ## Contribute
 
--- a/readme.md
+++ b/readme.md
@@ -810,6 +810,8 @@ It lets you define your own schema of what is and isn’t allowed.
   — hook based alternative
 * [`rehype-react`][github-rehype-react]
   — turn HTML into React elements
+* [`react-native-remark`][github-react-native-remark]
+  - remark on React Native
 
 ## Contribute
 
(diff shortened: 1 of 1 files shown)


ChristianMurphy (commented):
Thanks for sharing @imwithye!

Looking at https://www.npmjs.com/package/react-native-remark and https://github.com/imwithye/react-native-remark/

A few things:
1. The package isn't 1.0 (stable) yet, it is at 0.10.0 at the time of writing. I'm not comfortable recommending an unstable plugin (source: https://www.npmjs.com/package/react-native-remark)
2. There are no tests for the package to make sure it works https://github.com/imwithye/react-native-remark/blob/53e3d9fe325d9775c72cc44ca85392d57bc6f060/package.json#L11
3. Types that are required by the type definitions are listed as dev dependencies, which will cause some package managers (newer yarn and pnpm) to fail https://github.com/imwithye/react-native-remark/blob/53e3d9fe325d9775c72cc44ca85392d57bc6f060/package.json#L45
4. The docs don't describe what the option this plugin accepts are https://github.com/imwithye/react-native-remark/tree/main?tab=readme-ov-file#usage
5. The lint job doesn't lint check the code https://github.com/imwithye/react-native-remark/blob/53e3d9fe325d9775c72cc44ca85392d57bc6f060/.github/workflows/lint-pr.yml#L14
6. The release doesn't generate a provenance statement https://github.com/imwithye/react-native-remark/blob/53e3d9fe325d9775c72cc44ca85392d57bc6f060/.github/workflows/release.yml#L31 https://docs.npmjs.com/generating-provenance-statements
7. The type of the package is listed as module https://github.com/imwithye/react-native-remark/blob/53e3d9fe325d9775c72cc44ca85392d57bc6f060/package.json#L6 but the build is not configured to produce ESM https://github.com/imwithye/react-native-remark/blob/53e3d9fe325d9775c72cc44ca85392d57bc6f060/tsconfig.json#L4 for safely publishing it should be `node16`, `node18`, or `nodenext` https://www.typescriptlang.org/tsconfig/#module
8. There are no docs on what flavor(s) of markdown are supported. Commonmark is the default of react-markdown. I also see GFM as a dependency https://github.com/imwithye/react-native-remark/blob/53e3d9fe325d9775c72cc44ca85392d57bc6f060/package.json#L33, any others? (Directive, mdx, frontmatter, etc)

imwithye:
Thank you very much! Your advice worth a lot! I will update my repo based your suggestions

